PR #4, opened 2017-04-27
0.1/43 determina activities and design endpoints

charlesstrange2:
built lots of stuff. need more discipline

### DIFF
--- a/app/Voucher.php
+++ b/app/Voucher.php
@@ -22,7 +22,6 @@ class Voucher extends Model
      */
     protected $fillable = [
         'sponsor_id',
-        'assignee_id',
         'redeemer_id',
         'creditor_id',
         'code',
--- a/app/Voucher.php
+++ b/app/Voucher.php
@@ -47,4 +47,8 @@ class Voucher extends Model
         return $this->belongsTo(Trader::class);
     }
 
+    public static function findByCode($code)
+    {
+        return self::where('code', $code)->get()->first();
+    }
 }
--- a/app/Voucher.php
+++ b/app/Voucher.php
@@ -26,7 +26,7 @@ class Voucher extends Model
         'redeemer_id',
         'creditor_id',
         'code',
-        'currentstate' // SM_CONFIG looks at this.
+        'currentstate', // SM_CONFIG looks at this.
     ];
 
     /**
--- a/app/Voucher.php
+++ b/app/Voucher.php
@@ -10,6 +10,7 @@ class Voucher extends Model
 {
     use Statable; // import the state transition stuff.
     use SoftDeletes; // import soft delete.
+    protected $dates = ['deleted_at'];
 
     const HISTORY_MODEL = 'App\VoucherState'; // the related model to store the history
     const SM_CONFIG = 'Voucher'; // the SM graph to use
@@ -39,11 +40,6 @@ class Voucher extends Model
     public function sponsor()
     {
         return $this->belongsTo(Sponsor::class);
-    }
-
-    public function redeemer()
-    {
-        return $this->belongsTo(User::class);
     }
 
     public function creditor()
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -43,7 +43,7 @@ $factory->define(App\Market::class, function (Faker\Generator $faker) {
     return [
         'name' => $faker->company,
         'location' => $faker->postcode,
-        'sponsor_id' => $sponsor_id // a random sponsor
+        'sponsor_id' => $sponsor_id, // a random sponsor
     ];
 });
 
@@ -69,7 +69,6 @@ $factory->define(App\Voucher::class, function (Faker\Generator $faker) {
     return [
         'sponsor_id' => $sponsor_id, // a random sponsor
         'code' => $faker->ean8, // 8 digit barcode
-        'currentstate' => 'requested'
+        'currentstate' => 'requested',
     ];
 });
-
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -51,7 +51,7 @@ $factory->define(App\Trader::class, function (Faker\Generator $faker) {
 
     return [
         'name' => $faker->name,
-        'picURL' => null,
+        'pic_url' => null,
         'market_id' => null,
     ];
 });
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -43,7 +43,7 @@ $factory->define(App\Market::class, function (Faker\Generator $faker) {
     return [
         'name' => $faker->company,
         'location' => $faker->postcode,
-        'sponsor_id' => $sponsor_id
+        'sponsor_id' => $sponsor_id // a random sponsor
     ];
 });
 
@@ -67,7 +67,7 @@ $factory->define(App\Voucher::class, function (Faker\Generator $faker) {
     }
 
     return [
-        'sponsor_id' => $sponsor_id,
+        'sponsor_id' => $sponsor_id, // a random sponsor
         'code' => $faker->ean8, // 8 digit barcode
         'currentstate' => 'requested'
     ];
--- a/database/migrations/2017_04_21_162001_create_vouchers_table.php
+++ b/database/migrations/2017_04_21_162001_create_vouchers_table.php
@@ -15,20 +15,16 @@ class CreateVouchersTable extends Migration
     {
         Schema::create('vouchers', function (Blueprint $table) {
             $table->increments('id'); // a voucher *instance*.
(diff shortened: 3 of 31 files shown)


app/Voucher.php
```diff
@@ -41,14 +42,13 @@ public function sponsor()
...
     }
 
-    public function redeemer()
-    {
-        return $this->belongsTo(User::class);
-    }
-
     public function creditor()
     {
         return $this->belongsTo(Trader::class);
```
katjam: Why not stick to one or other of App\Trader vs Trader::class vs use App\Trader; Trader

database/factories/ModelFactory.php
```diff
@@ -43,15 +43,15 @@
     return [
         'name' => $faker->company,
         'location' => $faker->postcode,
-        'sponsor_id' => $sponsor_id
+        'sponsor_id' => $sponsor_id // a random sponsor
```
katjam: missed off the trailing ,

database/factories/ModelFactory.php
```diff
@@ -67,7 +67,7 @@
     }
 
     return [
-        'sponsor_id' => $sponsor_id,
+        'sponsor_id' => $sponsor_id, // a random sponsor
         'code' => $faker->ean8, // 8 digit barcode
         'currentstate' => 'requested'
```
katjam: missed off the trailing , 

I guess stick to one or the other standard
[
1,
2,
3,
] 
or 
[
1,
2,
3
]

database/migrations/2017_04_21_162001_create_vouchers_table.php
```diff
@@ -15,20 +15,16 @@ public function up()
     {
         Schema::create('vouchers', function (Blueprint $table) {
             $table->increments('id'); // a voucher *instance*.
-            $table->integer('assigner_id')->unsigned()->nullable(); // who was assigned this voucher.
-            $table->integer('redeemer_id')->unsigned()->nullable(); // who redeemed this with
+            $table->integer('assignee_id')->unsigned()->nullable(); // who was assigned this voucher.
```
katjam: Would be assigner I think - but should it be a user type - to make the schema and relations clearer?

database/migrations/2017_04_21_162001_create_vouchers_table.php
```diff
@@ -15,20 +15,16 @@ public function up()
...
             $table->integer('creditor_id')->unsigned()->nullable(); // who's owed money for this.
             $table->string('code', 32); // the actual voucher code.
             $table->string('currentstate', 24)->default('requested');
             $table->integer('sponsor_id')->unsigned();  // the organisation that sponsored this token (usually an LA).
             $table->timestamps();
+            $table->softDeletes();
         });
 
         Schema::table('vouchers', function (Blueprint $table) {
-            $table->foreign('redeemer_id')
```
katjam: Isn't assignee a foreign key?

database/migrations/2017_04_26_133415_create_trader_user_pivot_table.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class CreateTraderUserPivotTable extends Migration
```
katjam: Could this many to many be simplified if we had Customers and Traders - both extended User?

database/migrations/2017_04_27_141456_create_submission_states_table.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    public function up()
+    {
+        Schema::create('submission_states', function (Blueprint $table) {
+            $table->increments('id');
+            $table->string('transition');
+            $table->string('from')->nullable(); // What we were before this event.
+            $table->integer('user_id')->unsigned(); // a User who made this happen.
+            $table->integer('submission_id')->unsigned();
+            $table->string('to');   // What state we are now.
```
katjam: current?

database/migrations/2017_04_27_141453_create_submissions_table.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class CreateSubmissionsTable extends Migration
```
katjam: What is a submission? Can a voucher be submitted multiple times? Are they related?

database/migrations/2017_04_27_141550_create_voucher_submission_pivot_table.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class CreateVoucherSubmissionPivotTable extends Migration
```
katjam: Could this be simplified? Do we really need many to many?

Is a submission a transaction that submits possible multiple vouchers AND a voucher can be used in more than one submission?

database/migrations/2017_04_27_141453_create_submissions_table.php
charlesstrange2: Submission is like a scratchpad that has states "in progress", "awating review", "accepted", "rejected", "paid", "archived". supports
- users need to be able to construct a list to be dispatched. 
- user need to pickup where they left off. 
- users need to know whether thier submission has been paid or rejected (for some reason)
- the app needs to support progressive enhancement , so while the endpoint for submission might accept a fully realised client produced Submission datastructure it doesn't *have* to.

(though, there is a danger i'm overthinking this.)

database/migrations/2017_04_27_141550_create_voucher_submission_pivot_table.php
charlesstrange2: Submission is for one Trader the User adminsters.
A User records many Vouchers on a Sumbssion
Submissions have state (see above)
Vouchers MAY appear on many submissions BUT only sumbissions that are "accepted" can allocate thier associated Trader with the voucher. subsequent submissions will bounce it as "collected" by a Trader. 

attempts to cope with edge cases in collection i suppose.

database/migrations/2017_04_27_141456_create_submission_states_table.php
charlesstrange2: It's a record of a transition with a from and to state. Yes, the current state is fetched *from* the top record in a voucher's state stack, and is on a field called voucher->currentstate.

database/migrations/2017_04_26_133415_create_trader_user_pivot_table.php
charlesstrange2: A User is an identity that logs in to manipulate Traders and Vouchers. A User *might* in practice be a Trader but not always; in the systems context is "A person who makes submissions on behalf of one or more traders, and may *be* that trader". Users may well sublass or compose with roles at some point but right now I don't want to couple it too much.

database/migrations/2017_04_21_162001_create_vouchers_table.php
charlesstrange2: Ignore assignee. I'll pull it. It's supposed to be an ID for a family that the Children's Centres give vouchers to. I may even pull it if next weds determines that we cant store that data in here.

database/migrations/2017_04_21_162001_create_vouchers_table.php
charlesstrange2: nope it was a typo(!). see below about what asignee does (and why it might be pointless)

database/factories/ModelFactory.php
charlesstrange2: I naturally prefer the latter (becuase IE used to throw a fit with extra commas in JS); but Zend disagrees with me. i'll add commas as per that.

database/factories/ModelFactory.php
charlesstrange2: will add.

app/Voucher.php
charlesstrange2: any idea which one causes PHP less work when resolving?

katjam: Nope. Sorry - but the laravel docs seem to use App\Trader

database/migrations/2017_04_26_133415_create_trader_user_pivot_table.php
katjam: So right now your only users are Service Users?

Who are the entities that log in to use the API?

database/migrations/2017_04_27_141456_create_submission_states_table.php
katjam: ah ok.

PR #11, opened 2017-05-03
Exclude storage dir

katjam:


.travis.yml
```diff
@@ -18,7 +18,7 @@ before_deploy:
 deploy:
   provider: script
   skip_cleanup: true
-  script: rsync -r --delete-after --quiet $TRAVIS_BUILD_DIR/* $DEPLOY_USER@$DEPLOY_IP:/var/www/$DEPLOY_DIR
+  script: rsync -r --exclude 'storage' --delete-after --quiet $TRAVIS_BUILD_DIR/* $DEPLOY_USER@$DEPLOY_IP:/var/www/$DEPLOY_DIR
```
charlesstrange2: Shouldn't it be 
--exlude storage/

katjam: Don't think it matters - that will exclude dir called storage which is what I want. I think.

katjam: I maybe don't need the quotes.

PR #12, opened 2017-05-03
Remove all references to submissions

katjam:


katjam:
It is sad RIP submissions - they might make a comeback. Never know.

PR #14, opened 2017-05-04
0.1/76 improve seeding

katjam:
I have left some Todo in tests - but leaving that to one side for now.

This passes. I will get the voucher_states seeded by factory bit working another time. I think we are good with this for now - voucher states are hardcoded in the Voucher seeder anyway.


app/Voucher.php
```diff
@@ -22,20 +22,44 @@ class Voucher extends Model
...
     ];
 
+    /**
+     * Rules for validation.
+     *
+     * @var array
+     */
+    public static $rules = [
+        // Might need to add a 'sometimes' if any required fields can be absent from requests.
+        'trader_id' => ['numeric', 'exists:traders,id'],
```
charlesstrange2: Trader_id can be null; vouchers wont always be requested and need one?

katjam: That's fine - it's just the required ones that need a sometimes - if an update can be submitted without including their value

PR #20, opened 2017-05-10
0.2/83 model tests

katjam:
Not doing User tests new because we don't have one in use.

### DIFF
--- a/app/Trader.php
+++ b/app/Trader.php
@@ -29,11 +29,11 @@ class Trader extends Model
     protected $hidden = [
     ];
 
-    public function users()
-    {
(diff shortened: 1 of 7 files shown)


app/Trader.php
```diff
@@ -29,11 +29,11 @@ class Trader extends Model
     protected $hidden = [
     ];
 
-    public function users()
-    {
-        return $this->belongsToMany('App\Users');
-    }
-
+    /**
+     * Get the voucehrs belonging to this trader.
```
charlesstrange2: vouchers

PR #22, opened 2017-05-11
0.2/controller unit tests

katjam:
This includes the changes in my 0.2/tidy branch...
This is what's new here: https://github.com/neontribe/ARCVService/commit/4f92503454e252b54c48539107c67dcad3a33884

It is the only controller testing I propose for now because the dev on service is much later in the process. And we don't need more for now.


README.md
```diff
@@ -33,4 +50,4 @@ Union Road,
 Farnham, 
 Surrey, 
 England,
-GU9 7PT
\ No newline at end of file
+GU9 7PT
```
charlesstrange2: IDLE: Confused; why is this line different?

katjam: lol: no idea. Git is funny sometimes. It has moved down a significant number of lines so maybe git just playing it safe?

charlesstrange2:
either way, i think it's ok; apply squirrel-in-a-hat

PR #23, opened 2017-05-16
125 api login

katjam:
Might be overkill - but worth getting our heads round it for future?


katjam:
Yeah - i think we'll find we need a bit more config - but it's a start

PR #27, opened 2017-05-19
Altered state flow

charlesstrange2:


config/state-machine.php
```diff
@@ -15,8 +15,9 @@
             'printed',
             'dispatched',
             'allocated', // to families
-            'recorded',
-            'reimbursed',
+            'recorded', // submitted to shortlist
+            'payment_pending', // shortlist completed
+            'reimbursed', // paid out
```
katjam: I am liking your doc comments - otherwise this could be open to interpretation.

tests/Unit/Models/VoucherStateModelTest.php
```diff
@@ -46,6 +46,7 @@ public function testProgressVoucherState()
         // Should be 2 events, "ordered" and "printed".
         $this->assertEquals('printed', $voucher->currentstate);
         $this->assertEquals(2, $voucher->history()->count());
+
```
katjam: OK if you like the air.

config/state-machine.php
charlesstrange2: If you've got any suggestions for renaming states and the transitions to get to them there's still time to change it to be more context-y!

tests/Unit/Models/VoucherStateModelTest.php
charlesstrange2: I was going to alter the test. and then didnt. artefact removed.

PR #28, opened 2017-05-19
0.2/131 user trader api guards

katjam:


app/Http/Controllers/API/VoucherController.php
```diff
@@ -68,15 +68,8 @@ public function collect(Request $request)
             $trader = Trader::find(1);
         }
 
-        // Once we have implemented login...
+        // The user who performs the transition. Are we using this anywhere yet?
         $user = Auth::user();
```
charlesstrange2: He'll get associated with a transition in the voucher_states

katjam: But we don't use him here?

charlesstrange2:
reads...

You are correct

When a transition is enacted,
\App\Listeners\StateHistoryManager->postTransition sets teh
voucher_state.user_id to `auth()->id` (line 39).

PR #32, opened 2017-05-24
0.2/155 voucher history tests

katjam:


database/factories/ModelFactory.php
```diff
@@ -104,9 +105,52 @@
...
-
     return array_merge($voucher, [
         'currentstate' => 'requested',
     ]);
 });
+
+/**
+ * Voucher with currentstate allocated.
+ */
+$factory->defineAs(App\Voucher::class, 'allocated', function ($faker) use ($factory) {
```
katjam: To make writing tests easier - I can't instatiate the voucher inside the factory that creates it - so I have to add the states manually or do it in a Seeder or Test as I did for the Voucher Seeder. Did you also see that there seems to be a bug with the StateMachine or the way we are using it. VoucherStates don't seem to be recording from and source values.

PR #33, opened 2017-05-24
161 voucher policy

katjam:


app/Policies/VoucherPolicy.php
```diff
@@ -0,0 +1,74 @@
...
+     * @param  \App\Voucher  $voucher
+     * @return mixed
+     */
+    public function update(User $user, Voucher $voucher)
+    {
+        //
+    }
+
+    /**
+     * Determine whether the user can view the voucher.
```
charlesstrange2: view or collect?

katjam: lol can you tell which code block I copy pasted?

PR #34, opened 2017-05-25
0.2/183 api downloads

katjam:
No guarantees on this but in principle - it's all wired up.

Just need to write tests - and confirm. When I try to prod it manually I get Maatwebsite\Excel\Exceptions\LaravelExcelException: [ERROR]: Headers already sent

so... meh.

I am happy if you 2 are to merge in like this and then fix.


katjam:
Aww. Thanks both. :flower:

PR #38, opened 2017-06-01
0.2/203 the get payment button should progress all recorded vouchers to pending

charlesstrange2:
Changed collect function into a more general transition function.


app/Http/Controllers/API/VoucherController.php
```diff
@@ -40,6 +35,11 @@ public function collect(Request $request)
         */
 
         // Get out - no vouchers to process.
```
katjam: Get out - no transition specified.

PR #44, opened 2017-06-05
0.2/224 force write client key on reseed

katjam:


routes/web.php
```diff
@@ -12,7 +12,5 @@
 */
 
 Route::get('/', function () {
```
charlesstrange2: Should it be "/" and not something a bit more obscure?

katjam: I guess. It's just the dashboard, so I thought it would be easiest for Kat and Steph to find that way. It's not a www domain anyway. arcvservice-prealpha is pretty obscure to me.

katjam: Cross my heart - this feature will go or be adjusted to be on staging only.

PR #45, opened 2017-06-05
Single trader for Rolf

katjam:


database/seeds/UsersSeeder.php
```diff
@@ -54,5 +54,9 @@ public function run()
             }
             $user->traders()->sync($trade_ids);
         }
+
+        // So we reliably have one user with a single trader.
+        $trader = App\Trader::find(1);
+        $trader->traders()->sync([1]);
```
charlesstrange2: Am i correct in thinking this creates a user for the trader?

katjam: It syncs the trader to the user with the whole array so any existing others would be wiped out.

PR #49, opened 2017-06-07
0.2/261 date format

gingerbreadpup:


app/Http/Controllers/API/TraderController.php
```diff
@@ -114,7 +114,12 @@ public function showVouchers(Trader $trader)
                 ]);
             case 'application/json':
             default:
-                return response()->json($vouchers, 200);
+                return response()->json($vouchers->map(function($voucher){
```
katjam: Check syntax. You need a few more spaces and indents here.

You could also take the creation of the array out above the return to make it slightly less squished into closures.

katjam: Or rather than return Voucher objects wholly just get the bits you want and make a new array. This was a quick auto-implementation before we know the structure of the json needed by the app.

katjam: ... but this is fine. Apart from the missing spaces. :) :+1:

app/Http/Controllers/API/TraderController.php
```diff
@@ -114,7 +113,11 @@ public function showVouchers(Trader $trader)
...
+                    $newVoucher = $voucher->toArray();
+                    $newVoucher["updated_at"] = $voucher->updated_at->format('d-m-Y H:i.s');
+                    return $newVoucher;
+                }),200);
```
katjam: space before the 200.

katjam:
@gingerbreadpup @charlesstrange2 This good to go now?

PR #51, opened 2017-06-07
Set access_token timeout to an hour

charlesstrange2:


app/Providers/AuthServiceProvider.php
```diff
@@ -33,8 +33,7 @@ public function boot()
             $router->forTransientTokens();
         });
 
-        Passport::tokensExpireIn(Carbon::now()->addMinutes(10));
+        Passport::tokensExpireIn(Carbon::now()->addMinutes(60));
```
katjam: Ace!

PR #55, opened 2017-06-15
1.0/269 email copy

gingerbreadpup:
It might be better for me to just add my email copy into yours, Katja - let me know. Here's my PR anyway.

### DIFF
--- a/resources/views/api/emails/voucher_history_email.blade.php
+++ b/resources/views/api/emails/voucher_history_email.blade.php
@@ -2,6 +2,14 @@
 
 @section('content')
 
-    <p>Voucher History Email</p>
+    <p>Hi there,</p>
+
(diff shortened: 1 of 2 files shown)


katjam:
@gingerbreadpup I branched off mine. so this is good to go. Do you want the user name in there to say Hi Rolf? Or the trader name?

resources/views/api/emails/voucher_history_email.blade.php
```diff
@@ -2,6 +2,14 @@
 
 @section('content')
 
-    <p>Voucher History Email</p>
+    <p>Hi there,</p>
+
+    <p>You've requested a record of your voucher payment history, which is attached to this email.</p>
```
katjam: Technically the user is requesting the history on behalf of a trader. Should we re-word?

PR #60, opened 2017-06-22
1.0/email accessibility

alexalala:
let me know what you think / any improvements

https://trello.com/c/6gORjGC6/329-better-email-accessibility#

used http://www.accessible-email.org/ and everything now passes, the only problem it now has is that there are duplicate links in the footer (as are in the content), but I think it's important to keep them in.

🐰

### DIFF
--- a/resources/views/api/emails/voucher_history_email.blade.php
+++ b/resources/views/api/emails/voucher_history_email.blade.php
@@ -2,16 +2,18 @@
 
 @section('content')
 [xXx]
-	<h1 style="text-align: center;">Voucher payment history request</h1>
+	<div role="main">
+		<h1 style="text-align: center;">Voucher payment history request</h1>
 
-    <p>Hi {{ $user }},</p>
+	    <p>Hi {{ $user }},</p>
 
-    <p>You've requested a record of {{ $trader }}'s voucher payment history, which is attached to this email.</p>
-    <!-- Is this the right email> -->
-    <p>If you have any problems with opening or downloading the file attached, please <a href="mailto:arc@neontribe.co.uk">email arc@neontribe.co.uk</a>.</p>
+	    <p>You've requested a record of {{ $trader }}'s voucher payment history, which is attached to this email.</p>
 
-    <p>Thanks,<br>
-    Rose Vouchers<br>
-    <a href="http://www.alexandrarose.org.uk/">www.alexandrarose.org.uk</a></p>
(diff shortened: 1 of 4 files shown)


resources/views/api/emails/voucher_history_email.blade.php
```diff
@@ -2,14 +2,18 @@
...
+        <p>Hi {{ $user }},</p>
 
-    <p>Thanks,<br>
-    Rose Vouchers<br>
-    <a href="http://www.alexandrarose.org.uk/">www.alexandrarose.org.uk</a></p>
+        <p>You've requested a record of {{ $trader }}'s voucher payment history, which is attached to this email.</p>
+
+        <p>If you have any problems with opening or downloading the file attached, please <a href="mailto:arc@neontribe.co.uk">email arc@neontribe.co.uk</a>.</p>
+
+        <p>Thanks,<br>
```
katjam: Could move this and the main at the top to another tpl so that this one is Only the content of the mail.

BUT - as this is our only mail at the moment - fine. There is no other mail to share components with.

katjam:
@alexalala Looking at it again - I am not sure they should be in there twice.

Can you ask @gingerbreadpup what she thinks? Leave the [xXx] in anyway and maybe some notes if you don't resolve now.

Just noticed a bit of text in the layout that could go too 'Alexandra Rose Text'

resources/views/layouts/email.blade.php
```diff
@@ -16,8 +16,8 @@
 
         <div class="footer" style="margin-top: 40px; font-size: 13px; color: #999;">
             <p>Rose Vouchers text</p>
-            <p>For more information please go to <a href="http://www.alexandrarose.org.uk/">www.alexandrarose.org.uk</a> or email us at <a href="mailto:arc@neontribe.co.uk">arc@neontribe.co.uk</a>.</p>
-            <p><a href="http://www.alexandrarose.org.uk/privacy/">Privacy Policy</a></p>
+            <p>For more information please go to <a href="http://www.alexandrarose.org.uk/" role="link">www.alexandrarose.org.uk</a> or email us at <a href="mailto:arc@neontribe.co.uk">arc@neontribe.co.uk</a>.</p>
+            <p><a href="http://www.alexandrarose.org.uk/privacy/" role="link">Privacy Policy</a></p>
```
charlesstrange2: Until a service specific privacy policy is impmeneted, this will do. Card for future update (in the privacy policy card)

PR #61, opened 2017-06-22
Improve copy for email request

gingerbreadpup:
Will also make a PR to change this same text in fixtures.

### DIFF
--- a/app/Http/Controllers/API/TraderController.php
+++ b/app/Http/Controllers/API/TraderController.php
@@ -158,7 +158,7 @@ class TraderController extends Controller
 
         event(new VoucherHistoryEmailRequested(Auth::user(), $trader, $file));
 
-        return response()->json(['message' => '[xXx]The voucher history has been emailed.'], 202);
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/API/TraderController.php
```diff
@@ -158,7 +158,7 @@ public function emailVoucherHistory(Request $request, Trader $trader)
 
         event(new VoucherHistoryEmailRequested(Auth::user(), $trader, $file));
 
-        return response()->json(['message' => '[xXx]The voucher history has been emailed.'], 202);
+        return response()->json(['message' => '[xXx]Thanks. If you don\'t receive an email with your voucher history, please try again later.'], 202);
```
katjam: Can we take the xXx out?

katjam: Also note that @alexalala is working on a refactor of error messages at the moment.

alexalala: oh yeah, that was one of the ones I moved to my new api.php file! Will just amend it after this gets merged in

PR #65, opened 2017-06-22
api file added

alexalala:


resources/lang/en/api.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+return [
+	'messages'	=> [
+		'email_voucher_history' => '[xXx]Thanks. If you don\'t receive an email with your voucher history, please try again later.'
```
katjam: Great - why don't you axe the x and merge in master so you can un-do the previous PR Steph did with copy.

resources/lang/en/api.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+return [
+	'messages'	=> [
+		'email_voucher_history' => 'Thanks. If you don\'t receive an email with your voucher history, please try again later.'
```
katjam: @gingerbreadpup Worth noting that the history will be the trader's rather than the user's but I think for the Beta that is mostly the same person?

PR #67, opened 2017-06-23
1.0/353 printed voucher collect

katjam:
Called away on emergency for another project... did not have time for transition to collect not allowed tests.

They are on a trello card.


katjam:
I will do both before merge.

PR #76, opened 2017-07-17
1.1/454 auth for admins

katjam:
Charlie is off sick - so I am hoping you all can have a look over the admin authentication I have started to implement. It is blocking other dev in the sprint which ends this week.

I will flag for Charlies review when he gets back so it doesn't get forgotten but in the interim, if you can have look and point out anything glaringly dumb - would appreciate it!

Because we implemented the API user with the default one - I agreed with Charlie that we would create an admin one for service users. That's why I have needed to override some of the laravel defaults and places it expects to find things.

This is a rapid sprint, so also not massively covered by tests - but we expect to revisit the service portal in a future development round.


app/Http/Controllers/Service/AdminController.php
```diff
@@ -0,0 +1,30 @@
+<?php
```
katjam: Once we know what else the admin will need to do I might get rid of this controller.

katjam:
@karalan I did have trouble with endless loops at one point - so I will take another look.

I think / should be set to display the 'dashboard' if logged in and /login if not

PR #79, opened 2017-07-17
1.1/374 sponsor code and how to get payment message should be determined by the traders market

ob6160:


tests/Unit/Controllers/Api/TraderControllerTest.php
```diff
@@ -2,7 +2,6 @@
 
 namespace Tests\Unit\Controllers\Api;
 
-use Illuminate\Http\Request;
 use Tests\TestCase;
```
katjam: Not your fault but these should be alphabetical

tests/Unit/Controllers/Api/TraderControllerTest.php
```diff
@@ -54,6 +53,25 @@ protected function setUp()
         // Todo set some of the pended_at times to yesterday.
     }
 
+    /**
+     * Tests the for the api.trader.market route and controller.
+     */
+    public function testShowMarket()
```
katjam: Awesome. We aren;t doing complete coverage on this project but if we were - you'd want - can't see another trader's market etc.

tests/Unit/Models/MarketModelTest.php
```diff
@@ -25,6 +25,8 @@ public function testMarketIsCreatedWithExpectedAttributes()
         $this->assertInstanceOf(Market::class, $m);
         $this->assertNotNull($m->name);
         $this->assertNotNull($m->location);
+        $this->assertNotNull($m->payment_message);
+        $this->assertNotNull($m->sponsor_shortcode);
```
katjam: You have also added 2 methods to the Market model. Write basic tests for them. (See Trader and Voucher model tests for example)

database/factories/ModelFactory.php
```diff
@@ -44,6 +44,7 @@
         'name' => $faker->company,
         'location' => $faker->postcode,
         'sponsor_id' => $sponsor_id, // a random sponsor
+        'payment_message' => $faker->sentence($nbWords = 12, $variableNbWords = true),
```
ob6160: I might prefix this with "Placeholder Payment Message: " or something. What do you think?

app/Market.php
```diff
@@ -35,5 +40,35 @@ class Market extends Model
...
+    }
+
+    /**
+     * Get the sponsor belonging to this market.
+     *
+     * @return \App\Sponsor
+     */
+    public function sponsor()
+    {
+        return $this->hasOne(Sponsor::class, 'id', 'sponsor_id');
```
ob6160: I tried removing them but laravel didn't seem to magically detect them.

database/factories/ModelFactory.php
```diff
@@ -98,6 +99,39 @@
 });
 
 /**
+ * Trader with linked market and sponsor.
+ */
+$factory->defineAs(App\Trader::class, 'with_market_sponsor', function ($faker) use ($factory) {
```
ob6160: I've re-written this with your comments in mind.

tests/Unit/Models/MarketModelTest.php
ob6160: Added testMarketHasOneSponsor and testGetSponsorShortcodeAttribute

app/Market.php
katjam: Is it a belongs_to?

database/factories/ModelFactory.php
katjam: Yeah might be a good idea.

database/factories/ModelFactory.php
```diff
@@ -98,6 +99,31 @@
...
+    $sponsor_def = $factory->raw(App\Sponsor::class);
+    $sponsor_def = array_merge($sponsor_def, [
+        'id' => 100,
+    ]);
+    factory(App\Sponsor::class)->create($sponsor_def);
+
+    // Setup Market.
+    $market_def = $factory->raw(App\Market::class);
+    $market_def = array_merge($market_def, [
+        'sponsor_id' => 100,
```
katjam: I don't know - feels arbitrary to use 100. It should be ok to just use the 'next available id' and grab it with $market_def->id later.

Usually we use these specific factories to make things with fixed values on some attributes - like has a value vs. null. Or has a specific value. Like Trader from Mars factory(Trader::class, trader_from_mars)->create();

I may have led you down the garden path on this one... if this factory is only useful for your test right now AND we are making the relationship required later - we should probably ditch this. The array merge I meant just for the Trader... so it's getting all complicated.

tests/Unit/Models/MarketModelTest.php
katjam: Should be market belongs to sponsor.

ob6160: Modified this in latest commit.

database/migrations/2017_07_17_140109_add_payment_message_fieldToMarket.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
katjam: your migration file name should be snaked

katjam: A bit late but you can generate with command next time. https://laravel.com/docs/5.4/migrations#generating-migrations

 `php artisan make:migration add_votes_to_users_table --table=users`

tests/Unit/Models/MarketModelTest.php
```diff
@@ -2,19 +2,24 @@
...
     use DatabaseMigrations;
 
     protected $market;
+    protected $sponsor;
     protected function setUp()
     {
         parent::setUp();
         $this->market = factory(Market::class)->create();
+        $this->sponsor = factory(Sponsor::class)->create([
+            'id' => $this->market->sponsor_id = 10,
```
katjam: You should be able to get the sponsor. `$this->sponsor = $this->market->sponsor;`

tests/Unit/Models/MarketModelTest.php
```diff
@@ -25,11 +30,25 @@ public function testMarketIsCreatedWithExpectedAttributes()
         $this->assertInstanceOf(Market::class, $m);
         $this->assertNotNull($m->name);
         $this->assertNotNull($m->location);
+        $this->assertNotNull($m->payment_message);
+        $this->assertNotNull($m->sponsor_shortcode);
         $this->assertNotNull($m->sponsor_id);
         $this->assertInternalType('integer', $m->sponsor_id);
     }
```
katjam: double line

ob6160:
:fireworks: :champagne:

PR #82, opened 2017-07-19
1.1/458 we send a copy of requested payments as an email to anna

katjam:
Rough and ready!


app/Http/Controllers/API/TraderController.php
```diff
@@ -152,9 +151,11 @@ public function showVoucherHistory(Trader $trader)
      */
     public function emailVoucherHistory(Request $request, Trader $trader)
     {
+        $vouchers = $trader->vouchersConfirmed;
+        $title = 'A report containing voucher history.';
         // Request date string as dd-mm-yyyy
         $date = $request->submission_date ? $request->submission_date : null;
```
ob6160: There's a neat thing in PHP 7 called the null coalesce operator which makes this kind of assignment a lot easier.
This:
`$date = $request->submission_date ?? null;` would be the equivalent I believe. If it's against the code style for this project ignore me.

charlesstrange2: We target php7 and psr2 seems to permit conractions like this if they fit on one line; unless someone's codesniffer/ide plugins throws a fit, go a ahead. 

This line is fine how it is, though.

config/mail.php
```diff
@@ -60,6 +60,11 @@
         'name' => env('MAIL_FROM_NAME', 'Example'),
     ],
 
+    'to_admin' => [
+        'address' => env('MAIL_TO_ADMIN_ADDRESS', 'hello@example.com'),
```
charlesstrange2: aside: defaults when a user hasn't bothered with the relevant .env entries, for some reason;
Should these resolve to something we 
a) know should exist (postmaster@localhost?)
b) know should never exist (noreply@example.com)
c) something that demo's what we expect (voucheradmin@example.com) in case anyone's checking the code for help.
I'm tempted by b)

app/Http/Controllers/API/VoucherController.php
```diff
@@ -83,20 +86,58 @@ public function transition(Request $request)
...
                 $fail_codes[] = $voucher->code;
             }
         }
 
+        // If there are any confirmed ones... trigger the email
+        // event paymentRequestedEmail.
+        if (sizeof($vouchers_for_payment) > 0) {
+            // Todo If the email fails for some reason, the User will not be aware.
+            // The Vouchers will be transitioned but the ARC admin will not know.
+            $this->emailVoucherPaymentRequest($trader, $vouchers_for_payment);
```
charlesstrange2: This comment is correct - email is not a garuanteed medium. However, in the distant future we'd have an actiivities function in the portal - this card is marked as a stopgap until then.

PR #86, opened 2017-07-20
1.1/462 view live vouchers

gingerbreadpup:
This branch is branched off Katja's branch for adding admin vouchers... which is why there is some routing stuff in here. But if we were to merge this in then all we would need to do is add the form for the add voucher page and push that up separately... hope I haven't messed this up too much.

Anyway, this PR has some layout changes - I've put the sidebar code in an includes folder, and I've created pages for adding and viewing vouchers, which both have the dashboard interface.

I've also formatted the voucher data into a table.


gingerbreadpup:
Also - perhaps Anna doesn't need to see all of these columns - thoughts?

PR #87, opened 2017-07-20
1.1/444 add summary to body of email regarding voucher payments

ob6160:
Steph just needs to modify the copy text but I thought I'd make a PR to get a review on the code I've already written.


tests/Unit/Listeners/SendVoucherHistoryEmailTest.php
```diff
@@ -42,6 +43,8 @@ protected function setUp()
 
         // Set up voucher states.
         Auth::login($this->user);
+
+
```
gingerbreadpup: :bus:

app/Http/Controllers/API/TraderController.php
```diff
@@ -173,6 +179,20 @@ public function emailVoucherHistory(Request $request, Trader $trader)
         return response()->json(['message' => $response_text], 202);
     }
 
+    public function getMinMaxVoucherDates($vouchers) {
+        $sorted_vouchers = $vouchers->sortBy(function($voucher) {
+            return $voucher->paymentPendedOn->created_at->timestamp;
+        });
+
+        $max_date = $sorted_vouchers[0]->paymentPendedOn->created_at->format('d-m-Y');
```
charlesstrange2: sortby is ascendant small->big. so is max_date and min_date the wrong way round here. alternativly, sortByDesc() does it the other way.

ob6160: It turned out I had forgotten to run `->values()->all()` on the returned collection from the sort, before I did that the array elements were in an odd order.

app/Http/Controllers/API/TraderController.php
```diff
@@ -173,6 +179,20 @@ public function emailVoucherHistory(Request $request, Trader $trader)
         return response()->json(['message' => $response_text], 202);
     }
 
+    public function getMinMaxVoucherDates($vouchers) {
```
katjam: I think this should be in the Voucher class not the trader class. Mix Mx dates should be determined for a group of vouchers and has nothing to do with Trader.

ob6160: That makes sense, I'll move it.

app/Events/VoucherHistoryEmailRequested.php
```diff
@@ -18,17 +18,23 @@ class VoucherHistoryEmailRequested
...
+    public $max_date;
     public $file;
 
     /**
      * Create a new event instance.
      *
      * @return void
      */
-    public function __construct(User $user, Trader $trader, $file)
+    public function __construct(User $user, Trader $trader, $vouchers, $file, $date, $max_date = null)
```
katjam: You are pushing a lot of params in here. Is there a more modular way to do this?

ob6160: I could send the dates as an array..

katjam: I thought about that too, but not sure it helps. Is the User just the logged in user? Can you get the vouchers from the trader? Could you get the dates inside the listener instead? Those sort of thoughts... but leaving as is 'works' so up to you/ time.

ob6160: Well in the "VoucherPaymentRequested" event we pass in the vouchers directly, I was following the example set by that event. I think we could probably get the vouchers from the trader. I don't think it's necessarily worth the hassle.

katjam:
That was because it's a specific list of vouchers - 'submitted for
payment'. I think in yours it is all vouchers pended on a single day which
can be computed.

app/Events/VoucherHistoryEmailRequested.php
```diff
@@ -18,17 +18,22 @@ class VoucherHistoryEmailRequested
 
     public $user;
     public $trader;
+    public $vouchers;
```
katjam: Ditch this too?

tests/Unit/Listeners/SendVoucherHistoryEmailTest.php
```diff
@@ -42,23 +43,23 @@ protected function setUp()
...
             $v->applyTransition('dispatch');
             $v->applyTransition('allocate');
             $v->trader_id = 1;
             $v->applyTransition('collect');
+            $v->applyTransition('confirm');
         }
 
-        // Progress one to pending_payment.
-        $this->vouchers[0]->applyTransition('confirm');
+        \App\VoucherState::where('voucher_id', $this->vouchers[0]->id)
```
katjam: You don't need the \App bit here.

tests/Unit/Listeners/SendVoucherHistoryEmailTest.php
```diff
@@ -42,23 +43,23 @@ protected function setUp()
...
         Auth::login($this->user);
+
         foreach ($this->vouchers as $v) {
             $v->applyTransition('order');
             $v->applyTransition('print');
             $v->applyTransition('dispatch');
             $v->applyTransition('allocate');
             $v->trader_id = 1;
             $v->applyTransition('collect');
+            $v->applyTransition('confirm');
```
katjam: Is it better for the test if some are not confirmed?

ob6160: They all need to be confirmed otherwise `$voucher->paymentPendedOn` won't be defined.

katjam:
That's true. So the test would need to take into account that not all
vouchers have been pended. Just like real life. I am only looking at bits
here - so if you have the big picture and what you are doing it right...
keep it.

ob6160:
I was under the impression that the only ever case that we would be using this is with confirmed vouchers.
Chatted with Charlie about this yesterday, the reason why we need to be certain that vouchers provided to this method are confirmed is because in the sort function we depend upon "$voucher->paymentPendedOn->created_at->timestamp" existing.

tests/Unit/Listeners/SendVoucherHistoryEmailTest.php
katjam: What I meant was that there will be vouchers that are not confirmed in the database and the test needs to take into account that it is selecting confirmed ones from a bigger set.

But as I say - this comment is coming out of context so you probably know better.

ob6160: oh, I see. If we need to take into account that we're selecting confirmed ones we could take the vouchers from the `$trader->vouchersConfirmed` list in the test too then?

PR #93, opened 2017-07-21
1.1/446 admin can input batches of vouchers continued

charlesstrange2:
A little bit of UI massage (which may conflict with steph's work...) 
A bunch of validation, plus a custom validation rule
Found a named route conflict and fudged it
Fixed the success redirect (by explicitly redirecting)
Progressed the vouchers to "printed" whereupon they should redeem.
No tests :disappointed:


app/Http/Controllers/Service/Admin/VouchersController.php
```diff
@@ -22,7 +25,7 @@ class VouchersController extends Controller
      */
     public function index()
     {
-        $vouchers = Voucher::all();
+        $vouchers = DB::table('vouchers')->orderBy('id', 'desc')->paginate(50);
```
katjam: Isn't there... ? `Voucher::orderBy('id', 'desc')->paginate(50);`

app/Http/Controllers/Service/Admin/VouchersController.php
```diff
@@ -47,20 +50,63 @@ public function create()
      */
     public function storeBatch(Request $request)
     {
-        $new_vouchers = [];
+        $newVouchers = [];
+
+        $sponsorIds = \App\Sponsor::all()->pluck('id')->toArray();
```
katjam: `Sponsor::all()->pluck('id')->toArray();`

app/Http/Controllers/Service/Admin/VouchersController.php
```diff
@@ -47,20 +50,63 @@ public function create()
      */
     public function storeBatch(Request $request)
     {
-        $new_vouchers = [];
+        $newVouchers = [];
```
katjam: I think we snake_case our laravel php vars and camelCase our js vars usually but obviously not a deal breaker.

app/Http/Controllers/Service/Admin/VouchersController.php
```diff
@@ -47,20 +50,63 @@ public function create()
...
+
+        $sponsorIds = \App\Sponsor::all()->pluck('id')->toArray();
+
+        $voucherRules = [
+            'sponsor_id' => [
+                'required',
+                Rule::in($sponsorIds)
```
katjam: I *think* there is laravel magic that you don't need to get the sponsor ids.

`sponsor_id => ['required', 'exists:sponsors,id'];` should do the trick.

katjam: And you should store them it the Voucher class so they can be reused.
I think I already started there?

The start and end are part of this particular form, but not the sponsor_id.

app/Http/Controllers/Service/Admin/VouchersController.php
```diff
@@ -47,20 +50,63 @@ public function create()
...
-            $v->created_at = Carbon::now();
-            $v->updated_at = Carbon::now();
-            $new_vouchers[] = $v->attributesToArray();
+            $v->created_at = $nowTime;
+            $v->updated_at = $nowTime;
+            $newVouchers[] = $v->attributesToArray();
+        }
+        // batch insert.
+        // Todo : there's NO "this voucher already exists" checking!!
+        Voucher::insert($newVouchers);
```
katjam: Would `$vouchers = Voucher::insert($newVouchers);` get them? Or does that return some db object? Or a boolean?

app/Http/Controllers/Service/Admin/VouchersController.php
```diff
@@ -47,20 +50,63 @@ public function create()
...
+
+        $vouchers = Voucher::
```
katjam: Create a scope for it in the Voucher class? scopeRequestedAt($nowTime)

app/Http/Controllers/Service/Admin/VouchersController.php
```diff
@@ -47,20 +50,63 @@ public function create()
...
+        // batch progress
+        foreach ($vouchers as $voucher) {
+            $voucher->applyTransition('order');
+            $voucher->applyTransition('print');
+            // printed vouchers should now be redeemable.
         }
-        Voucher::insert($new_vouchers);
 
-        // Todo progress vouchers to allocated?
+        $notificationMsg = 'Created and activated '. $shortcode.$request['start'] .' to '. $shortcode.$request['end'];
```
katjam: Should be a message in lang/en/service.php or admin.php?

app/Http/Controllers/Service/Admin/VouchersController.php
charlesstrange2: will change

app/Http/Controllers/Service/Admin/VouchersController.php
charlesstrange2: Oh, OK. My bad.

app/Http/Controllers/Service/Admin/VouchersController.php
charlesstrange2: I thought so too, but my ide was more confident about this. Mind you php storm can be funny sometimes...

app/Http/Controllers/Service/Admin/VouchersController.php
charlesstrange2: Insert only gets the *last* updated value as a return value (not an array). Much sadness. It might be possible to get the current I'd, the  the final ID that way, but I've been bitten by non transactional inserts getting magically requeued amongst other db ops and the id's ending up non-contiguous.

app/Http/Controllers/Service/Admin/VouchersController.php
charlesstrange2: My reading if exists is that it checks if a column header exists in DB. Not sure it actually checks the values therein. A prepared to be wrong though!

PR #94, opened 2017-07-21
Update copy to include a couple of software suggestions for viewing a…

gingerbreadpup:
Didn't want to reel off too many program suggestions, so just went with Excel and Google Sheets...


resources/views/api/emails/voucher_history_email.blade.php
```diff
@@ -7,13 +7,13 @@
...
 
-        <p>[xXx] You've requested a copy of {{ $trader }}'s voucher payment records, which is attached to this email.</p>
+        <p>[xXx] You've requested a copy of {{ $trader }}'s voucher payment records for your reference, which is attached to this email.</p>
 
         <p>[xXx] The file includes payment records from {{ $date }} @isset($max_date)to {{ $max_date }}. @endisset</p>
         <p>[xXx] Total Vouchers: {{ count($vouchers) }}</p>
         <p>[xXx] Total Value: £{{ count($vouchers) }}</p>
 
-        <p>If you have any problems with opening or downloading the file attached, please <a href="mailto:arc@neontribe.co.uk">email arc@neontribe.co.uk</a>.</p>
+        <p>The attached file is best viewed through a spreadsheet program, such as Microsoft Excel or Google Sheets. If you have any problems with opening or downloading it, please email <a href="mailto:arc@neontribe.co.uk">arc@neontribe.co.uk</a>.</p>
```
charlesstrange2: No LibreOffice Calc?

gingerbreadpup: If you think it's actually likely that a market trader would use that, then sure. It's never really been on my radar so don't know how prevalent its use is!

charlesstrange2: Covering bases, I guess. Plus, for free, it's a quality product.

PR #95, opened 2017-07-21
1.1/481 notify arc admin when a duplicate has been entered

karalan:


karalan:
AFAIK yes, I have no idea what the trigger is yet as it's not been completed.

app/Mail/VoucherDuplicateEnteredEmail.php
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * Create a new message instance.
+     */
+
+    public function __construct($user, $trader, $voucher)
+    {
+        $this->user = $user->name;
+        $this->trader = $trader->name;
+        $this->voucher = $voucher;
+        $this->market = $trader->market;
```
katjam: We'll want the Market name or a suitable placeholder if empty.

charlesstrange2: i did 
$this->market = $trader->market ? $trader->market->name : 'no associated market'; 
on the others

charlesstrange2: @katjam I just updated this, alcan you review?

karalan: @katjam @charlesstrange2 I've updated and tests pass locally but Travis is shouting if there's an apostrophe in the copy.

resources/views/api/emails/voucher_duplicate_email_textonly.blade.php
```diff
@@ -0,0 +1,22 @@
...
+
+==============================================
+
+[xXx]
+Hi {{ config('mail.to_admin.name') }},
+
+{{ $user }} has tried to submit voucher
+{{ $vouchercode }} against
+{{ $trader }} of
+{{ $market->name ?? 'no associated market' }}'s account, however that voucher has already been submitted by another trader.
```
katjam: Please add to app/Mail/VoucherDuplicateEnteredEmail.php instead so you don't need to duplicate.

karalan: It doesn't work if you put it in there, because then the blade template is trying to get the name of something that isn't an object and it barfs. This is the only way I could get it to work?

katjam: You have to assign it like `$this>market = $this->market ? $this->market->name : 'no associated market';`

Then use `$market` in the tpl like `$trader` and `$user`.

katjam:
@karalan I pushed a fix. Tests pass - so I guess it's good to go.

PR #96, opened 2017-07-21
Fix object printing bug

charlesstrange2:


app/Mail/VoucherPaymentRequestEmail.php
```diff
@@ -28,7 +28,7 @@ public function __construct($user, $trader, $vouchers, $file)
         // Sending vouchers collection in case we want more than just count in email for copy.
         $this->vouchers = $vouchers;
         $this->file = $file;
-        $this->market = $trader->market;
+        $this->market = $trader->market->name;
```
katjam: We could probably ditch the market from here altogether. And get from Trader when it is used. Also we have not yet forced the requirement - so might be better as ternary with a default?

charlesstrange2: This may be why a test is failing.

katjam: In the sheet data - I think this:
```
         $data = [
 +            'report_title' => $title,
              'user' => Auth::user()->name,
              'trader' => $trader->name,
              // This is currently a nullable relation.
             'market' => $trader->market
                  ? $trader->market->name
                  : 'no associated market',
             'vouchers' => [],
          ];
```

PR #98, opened 2017-07-23
Xxx sweep and tweak copy

gingerbreadpup:


resources/lang/en/api.php
```diff
@@ -4,8 +4,8 @@
 
     'messages' => [
         'email_voucher_history' => 'Thanks. If you don\'t receive an email with your voucher records, please try again later.',
-        'email_voucher_history_date' => '[xXx] Thanks, we\'ve received your request for payment records from :date. If you don\'t receive an email with your voucher records, please try again later.',
-       'voucher_payment_requested' => '[xXx] Thanks. An email has been sent to let the team at Alexandra Rose that you have requested payment.',
+        'email_voucher_history_date' => 'Thanks, we\'ve received your request for payment records from :date. If you don\'t receive an email with your voucher records, please try again later.',
+       'voucher_payment_requested' => 'Thanks. The Alexandra Rose team have been notified of your payment request.',
```
katjam: One more space. My fault I think. :)

resources/views/api/emails/voucher_history_email.blade.php
```diff
@@ -3,15 +3,15 @@
 @section('content')
 
     <section role="main">
-        <h1 style="text-align: center;">[xXx] Voucher Payment Records</h1>
+        <h1>Voucher Payment Records</h1>
```
katjam: I am always unsure... is h1 for the site/ company name and h2 for page titles? No idea what happens in case of an email.

resources/views/api/emails/voucher_history_email.blade.php
```diff
@@ -3,15 +3,15 @@
...
         <p>Hi {{ $user }},</p>
 
-        <p>[xXx] You've requested a copy of {{ $trader }}'s voucher payment records for your reference, which is attached to this email.</p>
+        <p>You've requested a copy of {{ $trader }}'s voucher payment records for your reference, which is attached to this email.</p>
 
-        <p>[xXx] The file includes payment records from {{ $date }} @isset($max_date)to {{ $max_date }}. @endisset</p>
-        <p>[xXx] Total Vouchers: {{ count($vouchers) }}</p>
-        <p>[xXx] Total Value: £{{ count($vouchers) }}</p>
+        <p>The file includes payment records from {{ $date }} @isset($max_date)to {{ $max_date }}. @endisset</p>
+        <p>Total Vouchers: {{ count($vouchers) }}</p>
```
katjam: These will both be the same if vouchers are always £1. The repetition bothers me but I see why we do it.

resources/views/api/emails/voucher_history_email.blade.php
```diff
@@ -3,15 +3,15 @@
...
+        <p>Total Value: £{{ count($vouchers) }}</p>
 
         <p>The attached file is best viewed through a spreadsheet program such as Microsoft Excel, LibreOffice Calc or Google Sheets. If you have any problems with opening or downloading it, please email <a href="mailto:arc@neontribe.co.uk">arc@neontribe.co.uk</a>.</p>
```
katjam: The email should not be hardcoded here. We have .env vars for various ones. If we also need neontribe to be contactable - we should add that there.

resources/views/api/emails/voucher_history_email_textonly.blade.php
```diff
@@ -6,9 +6,9 @@
...
 You've requested a copy of {{ $trader }}'s voucher payment records for your reference, which is attached to this email.
 
-[xXx] The file includes payment records from {{ $date }}@isset($max_date) to {{ $max_date }}. @endisset
-[xXx] Total Vouchers: {{ count($vouchers) }}
-[xXx] Total Value: £{{ count($vouchers) }}
+The file includes payment records from {{ $date }}@isset($max_date) to {{ $max_date }}. @endisset
+Total Vouchers: {{ count($vouchers) }}
+Total Value: £{{ count($vouchers) }}
 
 The attached file is best viewed through a spreadsheet program such as Microsoft Excel, LibreOffice Calc or Google Sheets. If you have any problems with opening or downloading it, please email <a href="mailto:arc@neontribe.co.uk">arc@neontribe.co.uk</a>.
```
katjam: same comment re. email (I know it's not part of this PR but might as well be - or card as a bug)

resources/views/api/emails/voucher_payrequest_email.blade.php
```diff
@@ -3,9 +3,9 @@
 @section('content')
 
     <section role="main">
-        <h1 style="text-align: center;">Voucher Payment Request</h1>
+        <h1>Voucher Payment Request</h1>
 
-        <p>[xXx]Hi {{ config('mail.to_admin.name') }},</p>
+        <p>Hi {{ config('mail.to_admin.name') }},</p>
```
katjam: The password notification one uses Hello.

resources/views/api/emails/voucher_history_email.blade.php
gingerbreadpup: I think we can just use h1 as we have above. :+1:

resources/views/api/emails/voucher_payrequest_email.blade.php
gingerbreadpup: Since we use 'hi' everywhere else, I'll just change the one place where it uses 'hello' to 'hi' if that's okay.

resources/views/api/emails/voucher_history_email.blade.php
gingerbreadpup: I don't see any email vars in .env, should I make one and then put it in config/mail.php too?

resources/views/api/emails/voucher_history_email.blade.php
katjam: OK - consistency within the app is good enough for me.

resources/views/api/emails/voucher_history_email.blade.php
katjam: Yes - if you think it should go to other than admin (which will be set to something Anna gets on production).

We currently have MAIL_FROM_ADDRESS and MAIL_TO_ADMIN_ADDRESS

We don't generally direct mail back to devs from live apps, but might for the beta - so maybe MAIL_TO_DEVELOPER_TEAM might be appropriate if @charlesstrange2 thinks so too.

Yes - add to .env.example and config/mail.php

resources/views/api/emails/voucher_payrequest_email.blade.php
katjam: Hi sounds good to me.

charlesstrange2:
Hi (name) is what we use for momo, if I remember correctly. I also recall
it dodges a spamassasin rule that hello may trigger.

resources/views/api/emails/voucher_history_email.blade.php
```diff
@@ -3,15 +3,15 @@
 @section('content')
 
     <section role="main">
-        <h1 style="text-align: center;">[xXx] Voucher Payment Records</h1>
+        <h1>Voucher Payment Records</h1>
 
         <p>Hi {{ $user }},</p>
 
-        <p>[xXx] You've requested a copy of {{ $trader }}'s voucher payment records for your reference, which is attached to this email.</p>
+        <p>You've requested a copy of {{ $trader }}'s voucher payment records for your reference, which is attached to this email.</p>
```
charlesstrange2: "; this is attached " rather than ", which is attached "

resources/views/api/emails/voucher_history_email.blade.php
charlesstrange2: I like it, because it's super clear. Plus, we don't know if a voucher will always be £ 1

resources/views/api/emails/voucher_history_email.blade.php
charlesstrange2: I thought I made a card about this, but now can't find it. Either way, @katjam is right, this needs to be a variable,eventually. I have no objection to it being done now. It should be an arc "user support" address ideally, not technically a dev address,even if it eventually aliases to us.

resources/views/api/emails/voucher_history_email_textonly.blade.php
charlesstrange2: As aboce

resources/views/api/emails/voucher_payrequest_email.blade.php
charlesstrange2: Hi works, and avoids a spamassasin rule that hello sometimes triggers

charlesstrange2:
Ah, pants. Tests failing.

charlesstrange2:
Fix those as well, then :shipit:

resources/views/api/emails/voucher_history_email.blade.php
gingerbreadpup: Not keen on the repetition of 'this' but will make the change as requested!

charlesstrange2: "It's"  then?

.env.example
```diff
@@ -4,6 +4,7 @@ APP_KEY=
 APP_DEBUG=true
 APP_LOG_LEVEL=debug
 APP_URL=http://localhost:8080
+ARC_MARKET_URL=https://voucher-staging.alexandrarose.org.uk
```
katjam: I think we are keeping the .env.examples generic? But doesn't really matter

.env.example
```diff
@@ -29,6 +30,8 @@ MAIL_FROM_ADDRESS=from@example.com
 MAIL_FROM_NAME='Mailer Name'
 MAIL_TO_ADMIN_ADDRESS=to@example.com
 MAIL_TO_ADMIN_NAME='Admin Name'
+MAIL_TO_DEVELOPER_TEAM=arc@neontribe.co.uk
```
katjam: Again - doesn't matter but I would use support@example.com

config/mail.php
```diff
@@ -67,6 +67,20 @@
...
     |--------------------------------------------------------------------------
+    | Global default "To developer team" Address
+    |--------------------------------------------------------------------------
+    |
+    | An email address for contacting the development team
+    |
+    */
+
+    'to_developer' => [
+        'address' => env('MAIL_TO_DEVELOPER_TEAM', 'arc@neontribe.co.uk'),
```
katjam: support@example.com as default.

PR #101, opened 2017-07-24
1.1/511 improve payment request copy

gingerbreadpup:


database/factories/ModelFactory.php
```diff
@@ -55,7 +55,7 @@
         'name' => $faker->company,
         'location' => $faker->postcode,
         'sponsor_id' => $sponsor_id, // a random sponsor
-        'payment_message' => "Placeholder Payment Message: $payment_sentence",
+        'payment_message' => "Thanks, your payment request has been accepted. $payment_sentence",
```
katjam: This is a good candidate for the config/api.php message file. So we don't change in so many places.

katjam: Could even append to the accessor method for $market->payment_message

ob6160:
Cool - nice accessor method :shipit:

ob6160:
Does this still need the test/comment?

PR #102, opened 2017-07-24
1.1/271 better errors

katjam:
Phew! @alexalala and I finished this together. It goes with https://github.com/neontribe/ARCVMarket/pull/183

The Market one needs tests correcting.

All the copy needs checking and we need acceptance on the card with the expected messages for various situations.

I have not pushed with xXx - in case it gets on to Staging before demo. :)


resources/lang/en/api.php
```diff
@@ -3,12 +3,17 @@
 return [
 
     'messages' => [
+        'voucher_success' => '[xXx] Voucher is valid',
+        'batch_voucher_submit' => 'Thanks! Your queue has been successfully submitted. :success_amount voucher(s) accepted, :duplicate_amount duplicate(s) and :invalid_amount invalid.',
```
katjam: No it doesn't actually get displayed. I think there is a comment to that effect in the other file, but maybe put it here too.

gingerbreadpup:
Ollie has included my copy change requests in this card, so this PR can now be approved.

https://trello.com/c/xSHGbdNC/522-change-copy-for-invalid-voucher-code-error-message

PR #103, opened 2017-07-25
1.1/144 api password reset

katjam:


app/Http/Controllers/API/Auth/ForgotPasswordController.php
```diff
@@ -0,0 +1,49 @@
...
+    use SendsPasswordResetEmails;
+
+    /**
+     * Get the broker to be used during password reset.
+     *
+     * @return PasswordBroker
+     */
+    protected function broker()
+    {
+        return Password::broker('users');
```
katjam: @charlesstrange2 This is the broker I added... It's in the sends email form. So I guess you also needed it for the reset form?

charlesstrange2:
Yup, otherwise the passwords provider is defaulting to admins (hence the
422, no email address message)

PR #104, opened 2017-07-25
1.1/512 tests for add voucher batches

ob6160:


ob6160:
@charlesstrange2 I'm tempted just to extend "testStoreBatch" which already checks that vouchers were created so that it also checks that they are within the correct code range. Thoughts?

charlesstrange2:
I think the test should be distinct. "testStoreBatch" currently checks the UI response to the adding of vouchers. Rename it to "testStoreBatchSuccessMsg" or similar, then clone the vocher adding part into a new "testStoreBatch" test and continue from there.

tests/Unit/Controllers/Service/Admin/VoucherControllerTest.php
```diff
@@ -0,0 +1,117 @@
...
+        // Assert that eleven vouchers have been made and are in the expected state and land within the expected range.
+        $this->assertCount(11, $vouchers);
+        foreach($vouchers as $voucher) {
+            $this->assertEquals('printed', $voucher->currentstate);
+
+            // Assert that the voucher code starts with the Sponsor shortcode.
+            $this->assertStringStartsWith($shortcode, $voucher->code);
+
+            // Assert that the voucher code lands between 50 and 60 (our generated range).
+            $this->assertRegExp('/,*[5-6][0-9]/', $voucher->code);
```
charlesstrange2: technically this matches 50-69; can you narrow it?

ob6160: I'll look into narrowing after I've got the fix in for the number of generated vouchers in that other ticket.

charlesstrange2:
Yup, that makes sense

PR #120, opened 2017-08-02
Command to import legacy vouchers

charlesstrange2:


app/Console/Commands/LegacyImport.php
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'arc:legacyimport 
+                                {vouchers : file of voucher codes to import, one per line.}
+                                {shortcode : the sponsoring entity\'s shortcode (eg RVNT).}  
+                                {email : email address of AdminUser to act for.}
+                                ';
```
katjam: back 4 spaces?

app/Console/Commands/LegacyImport.php
```diff
@@ -0,0 +1,100 @@
...
+use Illuminate\Console\Command;
+use Auth;
+use \App\Sponsor;
+use \App\AdminUser;
+use \App\Voucher;
+
+class LegacyImport extends Command
+{
+
+    private $codes = [];
```
katjam: Is it worth specifying what format we need the codes/file to be in?

PR #123, opened 2018-07-23
Add the possibility to reject a recorded voucher back to allocated

charlesstrange2:


config/state-machine.php
```diff
@@ -50,6 +50,10 @@
                 'from' => ['printed','dispatched','allocated'],
                 'to' =>  'recorded',
             ],
+            'reject' => [
```
katjam: Why does reject move to allocated? What does it mean?

charlesstrange2: I chose reject, as the market app is rejecting it from the 'collect'ed vouchers list. The workflow then shunts it back to "allocated".

PR #126, opened 2018-07-25
1.2/968 write a voucher removal function so recorded ed vouchers can be rolled back include this in voucher controller

charlesstrange2:
Right, lets try this again.

System should now try to pick the correct reverse transition back to the original state based on the voucher *prior* state. Also, added more sensible messages in translations, and updated tests to cope.

### DIFF
--- a/app/Http/Controllers/API/VoucherController.php
+++ b/app/Http/Controllers/API/VoucherController.php
@@ -247,7 +247,7 @@ class VoucherController extends Controller
                 'message' => trans(
                     'api.messages.batch_voucher_submit',
                     [
-                    'success_amount' => count($responses['success']),
+                    'success_amount' => count($responses['success_add']),
                     'duplicate_amount' => count($responses['own_duplicate'])
                         + count($responses['other_duplicate']),
                     'invalid_amount' => count($responses['invalid']),
--- a/app/Http/Controllers/API/VoucherController.php
+++ b/app/Http/Controllers/API/VoucherController.php
@@ -88,13 +88,16 @@ class VoucherController extends Controller
             // can we?
             if ($voucher->transitionAllowed($transition)) {
                 // Some-what naive
-                $voucher->trader_id = $trader->id;
+
+                $voucher->trader_id = ($request['transition'] == 'reject')
+                    ? null
+                    : $trader->id;
 
                 // this saves the model too.
                 $voucher->applyTransition($transition);
 
                 // Success for this one.
-                if ($transition == $request['transition']) {
+                if ($request['transition'] != 'reject') {
                     $success_codes[] = $voucher->code;
                 } else {
                     $reject_codes[] = $voucher->code;
--- a/app/Http/Controllers/API/VoucherController.php
+++ b/app/Http/Controllers/API/VoucherController.php
@@ -3,11 +3,9 @@
 namespace App\Http\Controllers\API;
 
 use App\Events\VoucherPaymentRequested;
-use App\Http\Controllers\API\TraderController;
 use App\Http\Controllers\Controller;
 use App\Trader;
 use App\Voucher;
-use App\User;
 use Auth;
 use Carbon\Carbon;
 use Illuminate\Http\Request;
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/API/VoucherController.php
```diff
@@ -181,42 +201,56 @@ private function constructResponseMessage($responses)
...
                     break;
             }
         } else {
             return [
                 // Todo: This message needs work - but not this round.
-                'message' => trans('api.messages.batch_voucher_submit', [
+                'message' => trans(
+                    'api.messages.batch_voucher_submit',
+                    [
                     'success_amount' => count($responses['success']),
```
katjam: Does this need to be `success_add` and `success_reject`. I don't think the response has `success` key anymore.

app/Listeners/StateHistoryManager.php
```diff
@@ -33,7 +33,7 @@ public function postTransition(TransitionEvent $event)
 
         $model->history()->create([
             "transition" => $event->getTransition(),
-            "from" => "", // what the state was before.
+            "from" => $event->getState(), // what the state was before.
```
katjam: Will it be empty if there wasn't one? I think there was a trello card about this way back when - to implement when we started to care that the `from`s were not being recorded. I guess you can locate and annihilate?

app/Traits/Statable.php
```diff
@@ -36,6 +36,16 @@ public function getStateMachine()
...
     }
 
+    /**
+     * gets the prior state associated with the Stateable model.
+     *
+     * @return self::HISTORY_MODEL
+     */
+    public function getPriorState()
+    {
+        return $this->history()->get(null)->last();
```
katjam: Cool is the output predictable?

config/state-machine.php
```diff
@@ -50,6 +50,18 @@
                 'from' => ['printed','dispatched','allocated'],
                 'to' =>  'recorded',
             ],
+            'reject-to-printed' => [
```
katjam: So complicated! :)

app/Http/Controllers/API/VoucherController.php
```diff
@@ -115,10 +133,12 @@ public function transition(Request $request)
         // Currently we can
         //  'collect' - submit and
         //  'confirm' - request payment on.
-        $responses['success'] = $success_codes;
+        $responses['success_add'] = $success_codes;
+        $responses['success_reject'] = $reject_codes;
         $responses['own_duplicate'] = $own_duplicate_codes;
         $responses['other_duplicate'] = $other_duplicate_codes;
         $responses['invalid'] = $bad_codes;
+        $responses['failed_reject'] = $failed_rejects;
```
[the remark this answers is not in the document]
charlesstrange2: No, it;s an aweful pile of arrays that hold the exception states for each voucher transition.
Ideally i'd OO the whole thing - but time is short.

app/Http/Controllers/API/VoucherController.php
```diff
@@ -181,42 +201,56 @@ private function constructResponseMessage($responses)
...
         if ($total_submitted === 1) {
             switch ($error_type) {
-                case 'success':
+                case 'success_add':
                     return [
-                        'message' => trans('api.messages.voucher_success')
+                        'message' => trans('api.messages.voucher_success_add'),
                     ];
                     break;
+                case 'success_reject':
```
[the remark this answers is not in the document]
charlesstrange2: when a trader wants to reject a voucher inserted by error, yes.

app/Http/Controllers/API/VoucherController.php
charlesstrange2: Hmm. will check that...

app/Listeners/StateHistoryManager.php
charlesstrange2: yes, I think it pops out null.

app/Traits/Statable.php
charlesstrange2: yes, unlike `latest()` which can't tell between vouchers `created_at` in the same second, fetching the history orders by ID for `last()` to pop the end off. It shouldn't be a memory problem, in theory, as there shouldn't be very many transitions.

config/state-machine.php
charlesstrange2: I know. :cry: 

I thought "revert", but it isn't reverting, really. that implies we rollback the whole transaction and remove it, whereas i think we need a record of all transactions.

PR #133, opened 2018-07-31
wrapping faked name in e function to handle apostrophe

karalan:


tests/Unit/Listeners/SendVoucherHistoryEmailTest.php
```diff
@@ -91,7 +91,7 @@ public function testRequestVoucherHistoryEmail()
         $this->seeEmailWasSent()
             ->seeEmailTo($user->email)
             ->seeEmailSubject('Rose Voucher Payment Records')
-            ->seeEmailContains('Hi ' . $user->name)
+            ->seeEmailContains('Hi ' . e($user->name))
```
katjam: What does e do/ stand for and which package is the function from?

karalan: It's a laravel helper function https://laravel.com/docs/5.4/helpers#method-e

katjam: A nice - is it worth a comment or explicit use of htmlspecialchars? Could be hard to track down in future

karalan: I'll add a comment in = )

katjam: Sure - no big deal - was curious. Won't hurt. :)

PR #142, opened 2018-08-09
turn off indexing in the public .htaccess

charlesstrange2:


public/.htaccess
```diff
@@ -1,6 +1,6 @@
 <IfModule mod_rewrite.c>
     <IfModule mod_negotiation.c>
-        Options -MultiViews
+        Options -Indexes -MultiViews
```
katjam: Yay! :diamond_shape_with_a_dot_inside:

PR #145, opened 2018-08-10
adding middleware and tightening up cookie options

karalan:


app/Http/Kernel.php
```diff
@@ -17,6 +17,8 @@ class Kernel extends HttpKernel
         \Illuminate\Foundation\Http\Middleware\CheckForMaintenanceMode::class,
         \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
         \App\Http\Middleware\TrimStrings::class,
+        \App\Http\Middleware\XssHeader::class,
+        \App\Http\Middleware\ContentTypeOptionsHeader::class,
         \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
```
charlesstrange2: I think we need XFrameHeader in here too (is it automagical?)

karalan: Yeah my bad, is fixed.

PR #152, opened 2018-09-13
Hotfix reapplied for voucher code with spaces

karalan:
Just reapplying as fell out of the code base on rollup from 1.1 to 1.2


karalan:
@katjam It's just an exact copy of @charlesstrange2 's fix from October 2017, which I can't see has any tests. There is a card spawned from it for better voucher code validation and input sanitising.

PR #153, opened 2018-09-13
Hotfix reapplied for voucher code with spaces

karalan:
Just reapplying as fell out of the code base on rollup from 1.1 to 1.2

### DIFF
--- a/app/Http/Controllers/API/VoucherController.php
+++ b/app/Http/Controllers/API/VoucherController.php
@@ -54,7 +54,17 @@ class VoucherController extends Controller
             return response("Trader not found", 400);
         }
 
-        $uniqueVouchers = array_unique($request['vouchers']);
+        //remove spaces from voucher array values
+        $cleanedVouchers = array_map(
+            function($dirtycode) {
+                $badChars = [" ",];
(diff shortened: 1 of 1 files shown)


karalan:
@katjam yeah it's intentionally the same - it got put in 1.1 after 1.2 was rolled up so dropped out. I'm putting it in 1.3 here so it doesn't just happen again as we've already rolled up.

katjam:
Does it need to be in 1.2 as well then?

katjam:
What is the current release version?

karalan:
@katjam I thought this was the best way to do it, so it's been missed out of 1.2 which has been released, so as far as I'm aware will need to be released again now to fix this issue for current users.

I was then putting it into 1.3 separately to ensure it wasn't missed out, as last time it was hotfixed into 1.1 but after rollup was made for 1.2, so 1.1 was never merged into 1.2 after the fix. If that's the better way to do it for process and version control then happy to delete this one, merge the fix into 1.2 and then merge into 1.2 into 1.3.

katjam:
@karalan I think fine to leave it as it is. Was just a bit confusing. :) No need to waste time over it. Especially as there is now a conversation. :)

PR #158, opened 2018-09-28
1.3/1055 create model bundle and relationships

charlesstrange2:


app/Bundle.php
```diff
@@ -0,0 +1,89 @@
...
+        'entitlement',
+        'allocated_at',
+        'centre_id',
+        'family_id'
+    ];
+
+    protected $rules = [
+    ];
+
+    /**
```
katjam: Do you want all the boilerplate?

app/Bundle.php
```diff
@@ -0,0 +1,89 @@
...
+        'updated_at',
+        'allocated_at'  // When it was handed out.
+    ];
+
+    /**
+     * Return the Centre it was allocated to
+     *
+     * @return \Illuminate\Database\Eloquent\Relations\BelongsTo
+     */
+    public function centre()
```
katjam: If we don't use these relationships - do we want to define them? In this PR - they seem only to be used in tests - but I might be missing something.

database/factories/ModelFactory.php
```diff
@@ -176,6 +176,59 @@
...
+        ? $attributes['allocated_at']
+        : null
+    ;
+
+    // $centre is the centre we created the bundle at
+    $centre = null;
+
+    // if there is one supplied, find it
+    if (isset($attributes['centre_id'])) {
+        $centre = App\Centre::find($attributes['centre_id']);
```
katjam: Do we need to create one if not found or fallback to null?

charlesstrange2: centre_id cannot be null; it's the centre the bundle was made at.

katjam: I meant if the lookup fails - for bad data reasons... but I guess that will fall through to the next part with a `(!$centre)`

PR #159, opened 2018-10-01
1.3/1058 alter voucherstate to track userid and usertype

charlesstrange2:


database/migrations/2018_09_28_140458_add_user_type_to_voucher_states.php
```diff
@@ -0,0 +1,71 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('voucher_states', function (Blueprint $table) {
+            $table->string('user_type')
+                ->after('user_id')
+                ->default(""); // there needs to be a default, as this is not a null-able field.
```
katjam: Annoying - but I think it's better like this than to default to any actual type of user

database/migrations/2018_09_28_140458_add_user_type_to_voucher_states.php
```diff
@@ -0,0 +1,71 @@
...
+        });
+
+        // At time of creation, this is all the states we might see.
+        DB::update("UPDATE voucher_states SET user_type = 'AdminUser'
```
katjam: Just a thought - if the transitions are limited to certain user types - can this be an appended field instead? Or are you optimising for a future where different user types are performing same transition?

tests/Unit/Listeners/StateHistoryManagerTest.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        Auth::login($this->adminUser);
+        // create a voucher as printed
+
+        $v = factory(Voucher::class, 'requested')->create();
+        $v->applyTransition('order');
+        $v->applyTransition('print');
+        $v->applyTransition('dispatch');
+
+        $state = $v->history()->get(null)->last();
```
katjam: What doe s arg null do for `get()`?

Does `$v->history->last()` not return what you need?

tests/Unit/Listeners/StateHistoryManagerTest.php
```diff
@@ -0,0 +1,61 @@
...
+        $this->assertEquals(Auth::user()->id, $state->user_id);
+        $this->assertEquals(class_basename(Auth::user()), $state->user_type);
+
+        Auth::login($this->centreUser);
+        $v->applyTransition('allocate');
+
+        $state = $v->history()->get(null)->last();
+        $this->assertEquals(Auth::user()->id, $state->user_id);
+        $this->assertEquals(class_basename(Auth::user()), $state->user_type);
```
katjam: Can't decide whether hard coding the expected 'CentreUser' type would be better test than grabbing from classname in these cases?

PR #170, opened 2018-10-09
1.3/1044 create a voucher collection page

charlesstrange2:
It needs tests.
It needs a PUT route.

Can you let me know if the approach looks good?


app/Bundle.php
```diff
@@ -13,7 +17,7 @@ class Bundle extends Model
      */
     protected $fillable = [
         'entitlement',
-        'allocated_at',
+        'disbursed_at',
```
katjam: I like the new name.

app/Bundle.php
```diff
@@ -53,9 +57,100 @@ class Bundle extends Model
...
+     * @return array
+     */
+    private function alterVouchers(Collection $vouchers, array $codes, Bundle $bundle = null)
+    {
+        $errors = [];
+        $badCodes = array_diff($vouchers->pluck("code")->toArray(), $codes);
+
+        if (!isEmpty($badCodes)) {
+            //Stop! report the bad codes!
+            $errors["codes"] = (array_key_exists("badCodes", $errors))
```
katjam: Won't `$errors` always be `[]` here?

app/Bundle.php
```diff
@@ -53,9 +57,100 @@ class Bundle extends Model
...
+                ? array_merge($badCodes, $errors)
+                : $badCodes;
+        } else {
+            // Run all these.
+            $vouchers->each(
+                function (Voucher $voucher) use ($bundle, $errors) {
+                    try {
+                        $voucher->setBundle($bundle);
```
katjam: `$bundle` might be null here is it ok to set that?

app/Bundle.php
```diff
@@ -53,9 +57,100 @@ class Bundle extends Model
...
+        'disbursed_at'  // When it was handed out.
     ];
 
+    /**
+     * Refactored function that works out if we broke anything then adds vouchers.
+     *
+     * @param Collection $vouchers
+     * @param array $codes
+     * @param Bundle|null $bundle
+     * @return array
```
katjam: Is the return always an array of errors (which could be empty - signifying success or maybe that nothing happened)?

app/Http/Controllers/Store/BundleController.php
```diff
@@ -3,7 +3,9 @@
 namespace App\Http\Controllers\Store;
 
 use Auth;
+use App\Voucher;
```
katjam: I like to abc these... but not big deal

app/Http/Controllers/Store/BundleController.php
```diff
@@ -45,4 +47,58 @@ public function index()
...
+                        break;
+                    default:
+                        $messages[] = 'there was an unknown error';
+                        break;
+                }
+            }
+            // Spit the basic error messages back
+            return redirect()->route('store.registration.voucher-manager', ['registration' => $registration->id])
+                ->withInput()
+                ->with('error_message', ucfirst(join(', ', $messages) . '.'));
```
katjam: Neat.

app/Registration.php
```diff
@@ -69,6 +70,28 @@ public function centre()
...
+     * @return Model
+     */
+    public function currentBundle()
+    {
+        return $this->bundles()
+            ->where('disbursed_at', null)
+            ->orderBy('id', 'asc')
+            ->firstOrCreate(
+                [
+//                    "allocating_centre_id" =>  Auth::user()->centre,
```
katjam: Do you want this in or out?

app/Registration.php
```diff
@@ -69,6 +70,28 @@ public function centre()
...
+                    "registration_id" => $this->id,
+                ],
+                [
+                    "entitlement" => $this->family->entitlement,
```
katjam: Do you need to catch if $this->family is null?

app/Voucher.php
```diff
@@ -63,6 +64,46 @@ class Voucher extends Model
         'sponsor_id' => ['numeric', 'required', 'exists:sponsors,id'],
     ];
 
+
+    /**
+     * Voucher can clean it's codes.
+     * @param array $codes
+     * @return array
+     */
+    public static function cleanCodes(array $codes)
```
katjam: I guess you are assuming future may hold more bad chars? otherwise would strip spaces be more accurate name?

app/Voucher.php
```diff
@@ -63,6 +64,46 @@ class Voucher extends Model
...
+                $badChars = [" ",];
+                return str_replace($badChars, "", $code);
+            },
+            $codes
+        );
+    }
+
+    /**
+     * The methods we should call to remember to transition.
+     * Could probably be turned into an event listener?
```
katjam: Event listeners overrated. :wink:

app/Bundle.php
charlesstrange2: Yes. If it;s null, we've put a null bundle, so set the voucher's bundle_id to null (disassociate it)
voucher->setBundle also tries to rollback the transition

app/Bundle.php
charlesstrange2: If errors is empty, there are no errors, so success.

app/Http/Controllers/Store/BundleController.php
charlesstrange2: A bit lazy; I should probably be i18n'ing it. Problem for another time.

app/Registration.php
charlesstrange2: I'm undecided. I think I want it out. keeping it for if tests go horribly wrong

app/Registration.php
charlesstrange2: entitlement should never be null - it's a calculated attribute that returns an int (i think)

app/Voucher.php
charlesstrange2: Indeed, it would not suprise me if non-space codes or more mangling is required in the future.

app/Bundle.php
charlesstrange2: hmm. probably an artefact of me refactoring this function out of the syncVouchers. Will amend.

app/Registration.php
katjam: Just that you'd get a 500 if there is no family for bug reasons.

PR #172, opened 2018-10-09
redirect back to search page

alexalala:
https://trello.com/c/ulyhwe7V/1035-voucher-allocation-collection-screen-way-to-search-again

🐴

### DIFF
--- a/resources/views/store/manage_vouchers.blade.php
+++ b/resources/views/store/manage_vouchers.blade.php
@@ -164,7 +164,7 @@
                         </div>
                     </div>
                 </div>
-                <button class="short-button">Confirm pick up</button>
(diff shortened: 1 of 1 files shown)


resources/views/store/manage_vouchers.blade.php
```diff
@@ -164,7 +164,7 @@
                         </div>
                     </div>
                 </div>
-                <button class="short-button">Confirm pick up</button>
+                <button class="short-button" onclick="javascript:window.location.href='{{ URL::route("store.registration.index") }}'; return false;">Confirm pick up</button>
```
charlesstrange2: Since this is basically "yep, lets pretend we did that" at the moment, rather than actually doing a collection, this is OK

alexalala: yup!

PR #174, opened 2018-10-10
1.3/1044 creaate a voucher collection page add put route

charlesstrange2:
@katjam this a) fixes my (redundant) sync code and b) adds the route and a test.

### DIFF
--- a/tests/Unit/Routes/StoreRoutesTest.php
+++ b/tests/Unit/Routes/StoreRoutesTest.php
@@ -239,6 +239,59 @@ class StoreRoutesTest extends StoreTestCase
         }
     }
 
+    public function testVoucherManagerUpdateGate()
+    {
+        // Create a random registration with our centre.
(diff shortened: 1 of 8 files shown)


tests/Unit/Routes/StoreRoutesTest.php
```diff
@@ -239,6 +239,59 @@ public function testVoucherManageRouteGate()
...
+            ->call(
+                'PUT',
+                $put_route,
+                [] // should erase the vouchers.
+            );
+        $this->assertResponseStatus(302);
+
+        // I can call put on a page as a neighbor
+        Auth::logout();
+        $this->actingAs($this->neighborUser, 'store')
```
katjam: Neighbour user is same sponsor as the centre user which means they can act on each other's registrations?

charlesstrange2: Yep!

PR #176, opened 2018-10-11
1.3/1040 voucher allocation screen display a list of allocated vouchers e vv

alexalala:
https://trello.com/c/Gt92yRcp/1040-voucher-allocation-screen-display-a-list-of-allocated-vouchers-e-vv

🎢


database/seeds/DatabaseSeeder.php
```diff
@@ -40,6 +40,7 @@ public function devSeeds()
         $this->call(AdminUsersSeeder::class);
         $this->call(VouchersSeeder::class);
         $this->call(VoucherStatesSeeder::class);
+        $this->call(BundleSeeder::class);
```
charlesstrange2: We need to remember to append this to the UsabilitySeeds() file too, when we build Bobby Bundles and family.

PR #177, opened 2018-10-12
1.3/1068 add one or more vouchers to a bundle backend

charlesstrange2:
Ok! Should accept and guess a bunch of vouchers based on start and end!

### DIFF
--- a/app/Bundle.php
+++ b/app/Bundle.php
@@ -2,6 +2,7 @@
 
 namespace App;
 
+use Auth;
 use DB;
 use Illuminate\Support\Collection;
 use Illuminate\Database\Eloquent\Model;
@@ -62,6 +63,40 @@ class Bundle extends Model
     ];
 
     /**
+     * Adds voucher codes to a bundle
+     * @param $voucherCodes
+     * @return array
+     */
+    public function addVouchers($voucherCodes)
+    {
+        $self = $this;
+        $errors = [];
+
+        // Get current Codes for vouchers on the bundle (if any)
+        $currentCodes = $this->vouchers
+            ->pluck('code')
+            ->toArray();
+
+        // Calculate vouchers to add, so we don't try to add already bundled vouchers.
+        $addBundleCodes = array_diff($voucherCodes, $currentCodes);
+
+        // Find vouchers models that match codes to add.
+        $addVouchers = Voucher::whereIn('code', $addBundleCodes)->get();
+
+        // Add the voucher models to a specific bundle (this one)
+        $addErrors = $this->alterVouchers($addVouchers, $addBundleCodes, $self);
+
+        // if it threw any errors, merge those with the array.
+        if (!empty($addErrors)) {
+            $errors = array_merge_recursive($addErrors, $errors);
+        }
+
+        // an empty errors array means all good.
+        return $errors;
+    }
+
+
+    /**
      * Refactored function that works out if we broke anything then adds vouchers.
      *
      * @param Collection $vouchers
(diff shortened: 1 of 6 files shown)


app/Bundle.php
```diff
@@ -72,27 +107,28 @@ class Bundle extends Model
     public function alterVouchers(Collection $vouchers, array $codes = [], Bundle $bundle = null)
```
katjam: I'm not that keen on the $self/ $this reference of Bundle. Might simplify things if addVouchers took care of this... but if you are using to replace whole bundle elsewhere we need to keep?

app/Http/Controllers/Store/BundleController.php
```diff
@@ -69,19 +128,7 @@ public function update(StoreUpdateBundleRequest $request, Registration $registra
...
-                ->with('error_message', ucfirst(join(', ', $messages) . '.'));
-        } else {
-            // Otherwise, sure, return to the new view.
-            return redirect()->route('store.registration.voucher-manager', ['registration' => $registration->id])
-                ->with('message', 'Vouchers updated.');
-        }
+        return $this->redirectAfterRequest($bundle->syncVouchers($voucherCodes), $registration);
     }
 
     public function removeVoucherFromCurrentBundle(Registration $registration, Voucher $voucher)
```
katjam: Missing doc

app/Http/Controllers/Store/BundleController.php
```diff
@@ -91,54 +138,57 @@ public function removeVoucherFromCurrentBundle(Registration $registration, Vouch
...
             $errors = $bundle->alterVouchers(collect([$voucher]), [], null);
         } else {
             // Error it out (how did you get here?
             $errors["foreign"] = [$voucher->code];
         }
 
+        return $this->redirectAfterRequest($errors, $registration);
+    }
+
+    public function redirectAfterRequest($errors, $registration)
```
katjam: Missing doc

app/Voucher.php
```diff
@@ -81,6 +80,27 @@ function ($code) {
...
+     */
+    public static function splitShortcodeNumeric(string $code)
+    {
+        // Clean the code
+        $clean = self::cleanCodes([$code]);
+        $code = array_shift($clean);
+        // Init matches
+        $matches = [];
+        // split into named matche and return
+        if (preg_match("/^(?<shortcode>\D*)(?<number>\d+)$/", $code, $matches) == 1) {
```
katjam: Assuming this regex works.

app/Http/Controllers/Store/BundleController.php
```diff
@@ -51,7 +52,65 @@ public function index()
...
     /**
-     * Create OR Update a registrations current active bundle
+     * Does a single voucher.
+     *
+     * @param StoreAppendBundleRequest $request
+     * @param Registration $registration
+     * @return \Illuminate\Http\RedirectResponse
+     */
+
+    public function addVouchersToCurrentBundle(StoreAppendBundleRequest $request, Registration $registration)
```
katjam: Consider sharing some functionality with `/app/Http/Controllers/Service/Admin/VouchersController.php`

The diff is that submit comes with Sponsor id and numbers - but might be able to use some rules and `range()`

app/Voucher.php
```diff
@@ -202,17 +222,18 @@ public static function findByCodes($codes)
    *
    * @return array
    */
-  public static function getMinMaxVoucherDates(Collection $vouchers) {
```
katjam: Nothing new here, right?

app/Voucher.php
charlesstrange2: I did some manual testing on it. It uses fancy named group return values. Coild do with an auto test. Ill write one seperatly

app/Voucher.php
charlesstrange2: Nope, its just a bit of formatting

app/Bundle.php
charlesstrange2: I'm not either, but it happened organically while i was avoiding writing the same thing everywhere

PR #185, opened 2018-10-16
Docker tweaks

tobybatch:
These changes configure the tests to pass.  And add some helper utils.


.docker/Dockerfile
```diff
@@ -2,48 +2,70 @@ FROM php:7.0-apache
 
 RUN apt update && apt install -y \
         acl \
+        ccze \
```
charlesstrange2: Nice, but neccesary?

.docker/Dockerfile
```diff
@@ -2,48 +2,70 @@ FROM php:7.0-apache
...
-ENV APP_ENV local
 ENV APP_DEBUG true
+ENV APP_ENV local
 ENV APP_LOG_LEVEL debug
-ENV APP_URL http //0.0.0.0 80
-ENV APP_SEEDS Usability
+ENV APP_NAME docker
+ENV APP_SEEDS 'Dev'
+ENV APP_TIMEZONE Europe/London
+ENV APP_URL http://0.0.0.0 80
```
charlesstrange2: FYI, in production we use HTTPS, but bump it up from port 80.

.docker/install-tests.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+sudo apt-get -y install -y libxpm4 libxrender1 libgtk2.0-0 libnss3 libgconf-2-4 chromium xvfb gtk2-engines-pixbuf xfonts-cyrillic xfonts-100dpi xfonts-75dpi xfonts-base xfonts-scalable imagemagick x11-apps
```
charlesstrange2: We'd like to use Dusk, but never got round to making it work; chromium xvfb etc may be redundant in this build. We can keep 'em in for when we do need it, maybe?

.docker/vimrc
```diff
@@ -0,0 +1,69 @@
+colorscheme elflord
```
charlesstrange2: I have no idea what this whole config does to `vim` - personally I use `vi` due to a former live in  SCO OpenServer. Will running `vi` on ubuntu boot that or `vim` with this config in place?

charlesstrange2: Also, Docker containers are supposed to be single purpose, right - why do we need `vim` anyway?

PR #193, opened 2018-10-30
1.4/1094 list update pulse

Nikomus:
Looking for a bit of feedback on this one... It seems like overkill to change the bundle controller to track what the bundle used to contain, so I'm just adding the animation if the list contains some vouchers. This means it fires on remove events as well, but that is still an update to the list's content. Let me know what you think 🙂 

add

https://trello.com/c/DM4K9bmM/1094-add-visual-signifier-when-voucher-list-updates-espcially-if-list-closed-e-v

### DIFF
--- a/resources/views/store/manage_vouchers.blade.php
+++ b/resources/views/store/manage_vouchers.blade.php
@@ -112,8 +112,8 @@
                 </form>
                 <p class="center vh-spaced">You have added {{ $vouchers_amount }} vouchers</p>
                 <button id="collection-button" class="long-button">Go to voucher collection</button>
-                <div class="center">
-                    <span class="clickable-span">
+                <div class="center" id="vouchers-added">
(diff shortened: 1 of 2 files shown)


resources/views/store/manage_vouchers.blade.php
```diff
@@ -112,8 +112,8 @@
                 </form>
                 <p class="center vh-spaced">You have added {{ $vouchers_amount }} vouchers</p>
                 <button id="collection-button" class="long-button">Go to voucher collection</button>
-                <div class="center">
-                    <span class="clickable-span">
+                <div class="center" id="vouchers-added">
+                    <span class="clickable-span" >
```
alexalala: rogue space? 😈

Nikomus: Nooo

PR #195, opened 2018-10-31
1.4/1125 write back end submission process catch and validate

charlesstrange2:
https://trello.com/c/4ytgNNUB/1125-write-back-end-submission-process-catch-and-validate

### DIFF
--- a/app/Http/Controllers/Store/BundleController.php
+++ b/app/Http/Controllers/Store/BundleController.php
@@ -131,7 +131,7 @@ class BundleController extends Controller
             "collected_on"
         ]);
 
-        // refactored to NOT alter vouchers IF we don't mention them in form input.
+        // If we don't mention them in form input because we are updating status of existing bundle vouchers
         if ($request->exists('vouchers')) {
             $inputs['vouchers'] = $request->input('vouchers');
         }
--- a/app/Http/Controllers/Store/BundleController.php
+++ b/app/Http/Controllers/Store/BundleController.php
@@ -2,6 +2,7 @@
 
 namespace App\Http\Controllers\Store;
 
+use Log;
 use Auth;
 use App\Bundle;
 use App\Voucher;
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/Store/BundleController.php
```diff
@@ -126,17 +125,40 @@ public function addVouchersToCurrentBundle(StoreAppendBundleRequest $request, Re
...
-        // find our bundle.
+        $inputs = $request->only([
+            "collected_at",
+            "collected_by",
+            "collected_on"
+        ]);
+
+        // refactored to NOT alter vouchers IF we don't mention them in form input.
+        if ($request->exists('vouchers')) {
+            $inputs['vouchers'] = $request->input('vouchers');
```
katjam: What is the case where no vouchers in request - but don't want them cleared? When updating the status only? Maybe alter comment to clarify `IF we don't mention them in form input because we are updating status of existing bundle vouchers`

tests/Unit/Controllers/Store/BundleControllerTest.php
```diff
@@ -133,6 +133,77 @@ public function testICannotSubmitInvalidValuesToAppendVouchers()
...
+            $response->seeInSession('errors');
+            $errors = Session::get("errors")->get($field);
+
+            // Check our specific message is present
+            $this->assertContains($set['outcome'][$field], $errors);
+
+            // we follow that to the correct page;
+            $this->followRedirects()
+                ->seePageIs($route)
+                ->assertResponseStatus(200)
```
katjam: Should it be 200 if we are looking at errors?

tests/Unit/Controllers/Store/BundleControllerTest.php
```diff
@@ -239,4 +310,41 @@ public function testICanDeleteANamedVoucher()
...
+        $this->assertEquals(count($this->testCodes), $currentBundle->vouchers()->count());
+
+        // sync with only a single empty voucher string erases the vouchers.
+        $this->assertEquals(3, $currentBundle->vouchers()->count());
+
+        $this->actingAs($this->centreUser, 'store')
+            ->put($put_route, ['vouchers' => [''] ]);
+
+        $currentBundle->refresh();
+        //dd($currentBundle->vouchers->toArray());
```
katjam: remove `dd()`

app/Http/Controllers/Store/BundleController.php
charlesstrange2: Yep, will change is clearer

tests/Unit/Controllers/Store/BundleControllerTest.php
charlesstrange2: It's a redirect with form errors, so I'm hoping to land back on the page. At this point I can't check that the errors are in the html, because we haven't decided how/if to show them.

tests/Unit/Controllers/Store/BundleControllerTest.php
charlesstrange2: eek. fixed.

PR #196, opened 2018-10-31
Inserted form, so we can submit things.

charlesstrange2:
1129-write-collection-form-front-end-submission-process

### DIFF
--- a/resources/views/store/manage_vouchers.blade.php
+++ b/resources/views/store/manage_vouchers.blade.php
@@ -157,38 +157,51 @@
                 <div>
                     <p>There are {{ $vouchers_amount }} vouchers waiting for the family</p>
                 </div>
-                <div class="pick-up vh-spaced">
-                    <div>
-                        <i class="fa fa-user"></i>
+                <form method="post"
+                      action="{{ route('store.registration.vouchers.put', [ 'registration' => $registration->id ]) }}">
+                    {!! method_field('put') !!}
+                    {!! csrf_field() !!}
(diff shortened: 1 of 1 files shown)


charlesstrange2:
:cry: I think I broke the CSS for the buttons.

resources/views/store/manage_vouchers.blade.php
```diff
@@ -157,38 +157,50 @@
                 <div>
                     <p>There are {{ $vouchers_amount }} vouchers waiting for the family</p>
                 </div>
-                <div class="pick-up vh-spaced">
-                    <div>
-                        <i class="fa fa-user"></i>
+                <form method="post" name
```
Nikomus: Empty name attr here?

resources/views/store/manage_vouchers.blade.php
```diff
@@ -157,38 +157,50 @@
...
+                                    {-- Will need to be a list of all local centres eventually --}
+                                    <option value="{{ $centre->id }}">{{ $centre->name }}</option>
+                                </select>
+                            </div>
                         </div>
+                        <button class="short-button"
+                                type="submit">
+                            Confirm pick up
+                        </button>
+                        <button class="long-button"
```
katjam: Don't like having this button in the form - but if semantically allowed - I guess it'll work.

resources/views/store/manage_vouchers.blade.php
charlesstrange2: Oops! fixed

PR #199, opened 2018-11-07
1.4/1132 refactor to remove prepred allocated as voucherstates in favour of bundle as the disbursed entity

charlesstrange2:
Lots of deleteing, as prepared and allocated are no longer voucher states.
Bundle takes on the task of tracking allocated vouchers and it's own disbursement state.

### DIFF
--- a/database/migrations/2018_09_28_140458_add_user_type_to_voucher_states.php
+++ b/database/migrations/2018_09_28_140458_add_user_type_to_voucher_states.php
@@ -38,7 +38,6 @@ class AddUserTypeToVoucherStates extends Migration
                 'collect',
                 'reject-to-printed',
                 'reject-to-dispatched',
-                'reject-to-allocated',
                 'confirm'
             ) 
         ");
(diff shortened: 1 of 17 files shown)


database/migrations/2018_09_28_140458_add_user_type_to_voucher_states.php
```diff
@@ -47,12 +46,7 @@ public function up()
             UPDATE voucher_states SET user_type = 'CentreUser' 
             WHERE transition
             IN (
-                'bundle',
-                'disburse',
-                'unbundle-to-printed',
-                'unbundle-to-dispatched',
-                'lose',
-                'allocate'
+                'lose'
```
katjam: Is this right - just 'lose' vouchers need the CentreUser type now?

Where has this migration been run? Would it be safer to write a new one to update previous state?

charlesstrange2: It's only been run on staging. I'm prepared to scrub staging's db and reseed if this throws a clanger!

PR #200, opened 2018-11-07
1.4/1134 write backend functions to support bundle history

charlesstrange2:


app/Bundle.php
```diff
@@ -179,33 +178,63 @@ public function syncVouchers(array $voucherCodes)
...
+     * @return \Illuminate\Database\Eloquent\Relations\BelongsTo
+     */
+    public function disbursingCentre()
     {
         return $this->belongsTo(Centre::class);
     }
 
     /**
-     * Get the Registration this bundle is for
+     * Return the Carer it was disbursed to
```
katjam: Carer doesn't match CentreUser I think should be 'Centre User who disbersed the bundle'?

tests/Unit/Models/BundleModelTest.php
```diff
@@ -2,8 +2,10 @@
 
 namespace Tests;
 
+use App\Registration;
```
katjam: I like them alphabetical but meh.

PR #202, opened 2018-11-07
Docker tweaks (#185)

charlesstrange2:
* Small inprovements to the container setup

* Updated mem size in docker php

* Serving as artisan serve

* Removed debug echo


charlesstrange2:
@tobybatch made changes to his docker ticket on 1.3
I think we can add these to 1.4 by a straigh 1.3->1.4 PR.

PR #203, opened 2018-11-08
1.4/1126 write back end submission process 2

charlesstrange2:
Basic functionality in play.
https://trello.com/c/QoGCdPWa/1126-write-back-end-submission-process-update-bundle-dispersement
Could do with some further validation guards on letting users assign wacky values to the PUT update route. 
Also requires tests, see test card.
https://trello.com/c/gRCtBfV7/1138-write-tests-to-check-voucher-disbersement

### DIFF
--- a/app/Http/Controllers/Store/BundleController.php
+++ b/app/Http/Controllers/Store/BundleController.php
@@ -2,13 +2,17 @@
 
 namespace App\Http\Controllers\Store;
 
-use Auth;
 use App\Bundle;
+use App\Carer;
+use App\Centre;
 use App\Voucher;
 use App\Registration;
 use App\Http\Requests\StoreAppendBundleRequest;
 use App\Http\Requests\StoreUpdateBundleRequest;
 use App\Http\Controllers\Controller;
+use Auth;
+use Carbon\Carbon;
+use Log;
 
 class BundleController extends Controller
 {
@@ -113,7 +117,16 @@ class BundleController extends Controller
             }
         };
 
-        return $this->redirectAfterRequest($bundle->addVouchers($voucherCodes), $registration);
+        // return to page
+        $errors = $bundle->addVouchers($voucherCodes);
+
+        // Return to manager in all cases
+        $successRoute = $failRoute = route(
+            'store.registration.voucher-manager',
+            ['registration' => $registration->id]
+        );
+
+        return $this->redirectAfterRequest($errors, $successRoute, $failRoute);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Store/BundleController.php
```diff
@@ -162,10 +181,47 @@ function ($value) {
...
+
+                // Find and add the centre
+                $centre = Centre::findOrFail($inputs['collected_at']);
+                $bundle->disbursingCentre()->associate($centre);
+
+                // Add the current user as disbursingUser.
+                $bundle->disbursingUser()->associate(Auth::user());
+
+                // Store it.
+                $bundle->save();
```
RpprRoger: Large try/catches can hide errors which you don't expect. The expected errors _only_ occur on line 198 and 201. From what I can tell?

app/Http/Controllers/Store/BundleController.php
```diff
@@ -188,27 +244,33 @@ public function removeVoucherFromCurrentBundle(Registration $registration, Vouch
...
      */
-    public function redirectAfterRequest($errors, $registration)
+    public function redirectAfterRequest($errors, $successRoute, $failRoute)
     {
-        $route = route('store.registration.voucher-manager', ['registration' => $registration->id]);
         if (!empty($errors)) {
             // Assemble messages
             $messages = [];
             foreach ($errors as $type => $values) {
                 switch ($type) {
```
RpprRoger: so `errors["transactions"] == false` will still trigger this error message? That's a tad confusing.

app/Http/Controllers/Store/BundleController.php
```diff
@@ -162,10 +181,47 @@ function ($value) {
             $errors[] = $bundle->syncVouchers($voucherCodes);
         }
 
-        // TODO : disbursing bundle code to go here.
+        // Check we have values on our inputs; This should have been covered in validation...
```
RpprRoger: > This should have been covered in validation...

:)

app/Http/Controllers/Store/BundleController.php
```diff
@@ -162,10 +181,47 @@ function ($value) {
...
+        if (isset($inputs['collected_at']) &&
+            isset($inputs['collected_by']) &&
+            isset($inputs['collected_on'])
+        ) {
+            // Add the date;
+            $bundle->disbursed_at = Carbon::createFromFormat(
+                'Y-m-d',
+                $inputs['collected_on']
```
RpprRoger: Disbursed is the same as collected? I presume that makes sense from the data's pov but it's a bit confusing to read.

RpprRoger:
This code will work looks logical. I just have minor feedback

app/Http/Controllers/Store/BundleController.php
charlesstrange2: yeah. It should be, but i also thought i'd check they were actually filled here.

PR #205, opened 2018-11-09
Add box sizing and padding to search section

Nikomus:
- Amend the search page to have total height 100%, rather than 100% + 2rem

https://trello.com/c/vc0BgKnd/1137-duplicate-scroll-bars

### DIFF
--- a/public/store/css/main.css
+++ b/public/store/css/main.css
@@ -969,13 +969,13 @@ VOUCHER PAGES
 }
 
 @keyframes slidein {
-  from {
-    transform: translateX(150%);
-  }
+	from {
+		transform: translateX(150%);
+	}
 
-  to {
-    transform: translateX(0%);
-  }
+  	to {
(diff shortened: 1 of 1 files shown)


public/store/css/main.css
```diff
@@ -726,10 +726,9 @@ SEARCH PAGE
 */
 
 .search {
-	padding-top: 2rem;
-	background-color: #FFF;
-	width: 80%;
-	margin-left: 10%;
+    background-color: #FFF;
```
alexalala: tabbyyyyy

Nikomus: Whattt. I think this is because I restored the code from a stash and the existing tabs space were lumpy.

alexalala: ah probably 😂 my text editor is awful, so might have been me.. 👻

alexalala:
the keyframes still need some formatting luv <3

PR #206, opened 2018-11-09
1.4/1128 change the way we store carers

charlesstrange2:
Previously, we just erased the existing list of carers and rebuilt them whenever a change was made to the registration via the edit page. This naive solution was effective, but now we need to keep the id's constant as we record the collecting carer when they pickup vouchers at centres.

Thus we need to a) track actual new carers, b) amend any that changed c) permit the amending of the primary carer, d) remove any carers in the database that are not present in the data submitted.

### DIFF
--- a/app/Http/Controllers/Store/RegistrationController.php
+++ b/app/Http/Controllers/Store/RegistrationController.php
@@ -364,7 +364,7 @@ class RegistrationController extends Controller
         $registration = Registration::findOrFail($request->get('registration'));
 
         // Update primary carer.
-        $carerInput = (array) $request->get("carer");
+        $carerInput = (array) $request->get("pri_carer");
         $carerKey = key($carerInput);
         $carer = Carer::find($carerKey);
         if ($carer->name !== $carerInput[$carer->id]) {
--- a/app/Http/Controllers/Store/RegistrationController.php
+++ b/app/Http/Controllers/Store/RegistrationController.php
@@ -363,6 +363,10 @@ class RegistrationController extends Controller
         // Fetch Registration and Family
         $registration = Registration::findOrFail($request->get('registration'));
 
+        // NOTE: Following refactor where we needed to retain Carer ids.
+        // Possible that we might want to add flag to carer to ditinguish Main from Secondary,
+        // to simplify method below for sorting and updating carer entries.
+
         // Update primary carer.
         $carerInput = (array) $request->get("pri_carer");
         $carerKey = key($carerInput);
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Store/RegistrationController.php
```diff
@@ -358,18 +356,51 @@ function ($child) {
...
+        $carerKey = key($carerInput);
+        $carer = Carer::find($carerKey);
+        if ($carer->name !== $carerInput[$carer->id]) {
+            $carer->name = $carerInput[$carer->id];
+            $amendedCarers[] = $carer;
+        }
+
+        // Find secondary carers id's in the DB
+        $carersInput = (array) $request->get("sec_carers");
+        $carersKeys = $registration->family->carers->pluck("id")->toArray();
```
katjam: Assuming eloquent makes sure this will return an empty array rather than 500.

app/Http/Controllers/Store/RegistrationController.php
```diff
@@ -358,18 +356,51 @@ function ($child) {
...
+        // Update primary carer.
+        $carerInput = (array) $request->get("pri_carer");
+        $carerKey = key($carerInput);
+        $carer = Carer::find($carerKey);
+        if ($carer->name !== $carerInput[$carer->id]) {
+            $carer->name = $carerInput[$carer->id];
+            $amendedCarers[] = $carer;
+        }
+
+        // Find secondary carers id's in the DB
```
katjam: Would there be any point in adding a flag to carer as Main? So we have a distinct group for each - and can ensure that we always have a Main carer?

Would save all the manipulation with array keys below.

app/Http/Controllers/Store/RegistrationController.php
```diff
@@ -358,18 +356,51 @@ function ($child) {
...
+
+        // roll though those and amend them if necessary.
+        foreach ($carers as $carer) {
+            if ($carer->name !== $carersInput[$carer->id]) {
+                $carer->name = $carersInput[$carer->id];
+                $amendedCarers[] = $carer;
+            }
+        }
+
+        // Create mew carers
```
katjam: typo new

app/Http/Requests/StoreUpdateRegistrationRequest.php
```diff
@@ -33,9 +33,11 @@ public function rules()
          */
         $rules = [
             // MUST be present; MUST be a not-null string
-            'carer' => 'required|string',
+            'pri_carer.*' => 'required|string',
             // MAY be present; MUST be a not-null string
-            'carers.*' => 'string',
+            'sec_carers.*' => 'string',
+            // MAY be present; MUST be a not-null string
+            'new_carers.*' => 'string',
```
katjam: Feels like there might be a better model that doesn't need `new_carers`

PR #207, opened 2018-11-09
styling selector and now reverses dates properly

alexalala:
https://trello.com/c/qyu1mjXn/1099-voucher-allocation-collection-screen-full-collection-history-eee-vvv

tacked onto this ticket, was a scoping problem so removed needless variable.

### DIFF
--- a/public/store/css/main.css
+++ b/public/store/css/main.css
@@ -829,12 +829,6 @@ HISTORY PAGE
 	text-align: center;
 }
 
-.history > p {
-	color: red;
-	margin: 5%;
-	font-size: 1.25rem;
-}
-
 .history table {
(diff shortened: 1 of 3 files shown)


public/store/css/main.css
```diff
@@ -829,7 +829,7 @@ HISTORY PAGE
 	text-align: center;
 }
 
-.history  > div > p {
+.history > p {
```
Nikomus: Should it be #d2232a, rather than red?

Not going to worry about it now at all, but I'm not a huge fan of child selectors from page-level classes. It can make it difficult to change things later.

alexalala: 👍 you right, fixed for this instance

PR #210, opened 2018-11-12
1.4/1124 edit page display fix 2

Nikomus:
This ended up being a lot less straightforward than I hoped...
- Shuffled the forms around with the remove button into a third column
- Extra jQuery to modify heights

arcv-store test_registrations_21_edit ipad


alexalala:
screenshot 2018-11-12 at 10 40 11
button isn't quite vertically centred within parent container I don't think. Just tried centering the div using flex and it seems to work alright!

alexalala:
Still working great ⛵️ 😸

PR #212, opened 2018-11-12
1.4/1150 bug prevent disbursing of an empty bundle

charlesstrange2:


app/Http/Controllers/Store/BundleController.php
```diff
@@ -186,39 +186,48 @@ function ($value) {
...
+                    $bundle->disbursingUser()->associate(Auth::user());
+
+                    // Store it.
+                    $bundle->save();
+
+                } catch (\Exception $e) {
+                    // Fires if finOrFail() or save() breaks
+                    // Log that error by hand
+                    Log::error('Bad transaction for ' . __CLASS__ . '@' . __METHOD__ . ' by service user ' . Auth::id());
+                    Log::error($e->getTraceAsString());
```
katjam: Is this trace log left in intentionally?

app/Http/Controllers/Store/BundleController.php
```diff
@@ -186,39 +186,48 @@ function ($value) {
...
-                $errors['transaction'] = true;
-            }
 
-            // Return to Index as we've disbursed, and user may want to search
-            $successRoute =  route(
-                'store.registration.index'
-            );
+            // Check there are actual vouchers to disburse, or this is a bit.
+            if ($bundle->vouchers->count() == 0) {
+
```
katjam: Extra blank lines and double quoted key strings not in keeping with rest of code style.

app/Http/Controllers/Store/BundleController.php
```diff
@@ -282,6 +291,11 @@ public function redirectAfterRequest($errors, $successRoute, $failRoute)
                     case "foreign":
                         $messages[] = "Action denied on a foreign voucher: " . join(', ', $values);
                         break;
+                    case "empty":
```
katjam: There might be an alternative to relying on keyed bools for error string generation. But happy to keep going with this for now.

PR #217, opened 2018-11-19
Make sure sponsor ids of CC User's centres match

Nikomus:
- Adds two extra centres to the CC User that definitely have the same sponsor id as the first one.
- It might be nicer to check how many there are already, and only make more if we need to, and attach the lot in any case?

General Laravel guidance appreciated 🙂 

https://trello.com/c/O2kvdGXJ/1153-centre-user-switching-httpexception

image

### DIFF
--- a/database/seeds/CentreUsersSeeder.php
+++ b/database/seeds/CentreUsersSeeder.php
@@ -18,10 +18,22 @@ class CentreUsersSeeder extends Seeder
             "password" => bcrypt('store_pass'),
             "role" => "centre_user",
         ]);
+
+        // Attach an initial centre
         $user1->centres()->attach([
-            1 => ['homeCentre' => true],
-            2 => ['homeCentre' => false],
-            3 => ['homeCentre' => false],
+            1 => ['homeCentre' => true]
+        ]);
+
+        // Get the first centre's sponsor and make two more centres with the same sponsor
+        $sponsor_id = $user1->centres()->first()->sponsor()->first()->id;
+
+        $local_centre1 = factory(App\Centre::class)->create(['sponsor_id' => $sponsor_id]);
+        $local_centre2 = factory(App\Centre::class)->create(['sponsor_id' => $sponsor_id]);
+
+        // Attach the extra centres
+        $user1->centres()->attach([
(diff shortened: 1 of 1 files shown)


database/seeds/CentreUsersSeeder.php
```diff
@@ -18,10 +18,22 @@ public function run()
...
+
+        // Attach an initial centre
+        $user1->centres()->attach([
+            1 => ['homeCentre' => true]
+        ]);
+
+        // Get the first centre's sponsor and make two more centres with the same sponsor
+        $sponsor_id = $user1->centres()->first()->sponsor()->first()->id;
+
+        $local_centre1 = factory(App\Centre::class)->create(['sponsor_id' => $sponsor_id]);
```
katjam: Looks good. You can also do this:
`$local_centres = factory(App\Centre::class, 2)->create(['sponsor_id' => $sponsor_id]);`

and access with:
`$local_centres[0]` and `$local_centres[1]`

Nikomus: @katjam This is much neater, I have updated to this.

katjam:
@Nikomus Was more for info for another more complicated case in future... but great!

database/seeds/CentreUsersSeeder.php
```diff
@@ -18,10 +18,21 @@ public function run()
...
             "role" => "centre_user",
         ]);
+
+        // Attach an initial centre
+        $user1->centres()->attach([
+            1 => ['homeCentre' => true]
+        ]);
+
+        // Get the first centre's sponsor and make two more centres with the same sponsor
+        $sponsor_id = $user1->centres()->first()->sponsor()->first()->id;
```
charlesstrange2: `$sponsor_id = $user1->centres()->first()`... 

- Up to this point because of lazy loading we *may* be dealing with calls to a database, depending on what state $user1 was in. `first()` is like `get()`, but limited to a single record. That gets turned into a single `Centre` model object.
- A `Centre` only has 1 sponsor, so you don't need ...`sponsor()->first()->id`; you can get away with `sponsor->id` as `sponsor` is a property of the model, and `id` is a property of that.

PR #221, opened 2018-11-23
1.4/1160 reset password red error message second crack

charlesstrange2:
different approach to @Nikomus' branch. I think the issue is that on password reset we *by default* log a person in. This lead the middleware to throw them at the dashboard page.

However, the use of `login($user)` is done in the wrong context, the "default" of web from config/auth['defaults'] and it gets screwed up looking at teh wrong table for a second befire adopting the write context and trying again.

### DIFF
--- a/resources/views/store/auth/login.blade.php
+++ b/resources/views/store/auth/login.blade.php
@@ -17,8 +17,13 @@
         <div class="login-container">
             <h2>Log In</h2>
             @if ($errors->has('error_message'))
-                <div class="alert alert-danger">
+                <div class="alert alert-warning">
                     <strong>{{ $errors->first('error_message') }}</strong>
+                </div>
+            @endif
+            @if (session('status'))
+                <div class="alert alert-info">
+                    <strong>{{ session('status') }}</strong>
                 </div>
             @endif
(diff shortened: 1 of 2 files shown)


resources/views/store/auth/login.blade.php
```diff
@@ -17,10 +17,15 @@
         <div class="login-container">
             <h2>Log In</h2>
             @if ($errors->has('error_message'))
-                <div class="alert alert-danger">
+                <div class="alert alert-warning">
                     <strong>{{ $errors->first('error_message') }}</strong>
                 </div>
             @endif
+            @if (session('status'))
```
katjam: What messages are you seeing in status? Just curious.

charlesstrange2: Auth returns "Your password has been reset!"

charlesstrange2:
As for setting the correct default on a per domain basis, I'm not sure. My thoughts go to having a dynamic configuration property for the "deafults" variable in config/auth.php. It is, however, a bit of a hack!

resources/views/store/auth/login.blade.php
katjam: Awesome.

PR #225, opened 2018-11-26
Discovered that we had too much randomness for our concrete users.

charlesstrange2:
Reduced that by fixing some centres into the right sponsors.

### DIFF
--- a/database/seeds/CentreUsersSeeder.php
+++ b/database/seeds/CentreUsersSeeder.php
@@ -25,7 +25,7 @@ class CentreUsersSeeder extends Seeder
         ]);
 
         // Get the first centre's sponsor and make two more centres with the same sponsor
-        $sponsor_id = $user1->centres()->first()->sponsor()->first()->id;
+        $sponsor_id = $user1->centres()->first()->sponsor->id;
 
         $local_centres = factory(App\Centre::class, 2)->create(['sponsor_id' => $sponsor_id]);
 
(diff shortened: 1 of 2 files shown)


database/seeds/CentreUsersSeeder.php
```diff
@@ -25,7 +25,7 @@ public function run()
         ]);
 
         // Get the first centre's sponsor and make two more centres with the same sponsor
-        $sponsor_id = $user1->centres()->first()->sponsor()->first()->id;
+        $sponsor_id = $user1->centres()->first()->sponsor->id;
```
katjam: Makes more sense

database/seeds/CentreUsersSeeder.php
```diff
@@ -50,7 +50,7 @@ public function run()
             "password" => bcrypt('store_pass'),
             "role" => "foodmatters_user",
         ]);
-        $user3->centres()->attach(1, ['homeCentre' => true]);
+        $user3->centres()->attach(2, ['homeCentre' => true]);
```
katjam: I'm guessing there is a good reason to go for second centre.

database/seeds/CentresSeeder.php
```diff
@@ -11,15 +11,21 @@ class CentresSeeder extends Seeder
...
-        $centres = factory(App\Centre::class, 4)->create();
+        $centres = factory(App\Centre::class, 3)->create();
 
         // Grab one and change print pref to individual and another to collection.
         $centres[0]->print_pref = 'collection';
         $centres[0]->save();
         $centres[1]->print_pref = 'individual';
         $centres[1]->save();
 
+
```
katjam: lots of extra whiteness

database/seeds/CentreUsersSeeder.php
charlesstrange2: Yes, the second centre gets a different print_pref, and at least one user ought to have it so karl cna test.

PR #228, opened 2018-11-28
Route doesn't take raw paths. remove it and just redirect.

charlesstrange2:


charlesstrange2:
Ta @Nikomus; You may as well mark this as tested as well?

PR #230, opened 2018-11-29
1.4/1180 a user cannot alter vouchers on a submitted disbursed bundle

charlesstrange2:
The object of this is to prevent the transfer of vouchers from a disbursed bundle to another bundle.

### DIFF
--- a/app/Bundle.php
+++ b/app/Bundle.php
@@ -118,11 +118,13 @@ class Bundle extends Model
 
         // Try to run the vouchers we know are in the DB
         $vouchers->each(
-            function (Voucher $voucher) use ($bundle, $errors) {
+            // Passing a pointer to $errors using "&", so we can change it, inside the loop.
+            // Otherwise the variable is immutable.
+            function (Voucher $voucher) use ($bundle, &$errors) {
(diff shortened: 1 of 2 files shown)


app/Bundle.php
```diff
@@ -116,10 +116,19 @@ public function alterVouchers(Collection $vouchers, array $codes = [], Bundle $b
...
-                $voucher->bundle()->associate($bundle)->save();
+            // Passing a pointer to $errors using "&", so we can change it, inside the loop.
+            // Otherwise the variable is immutable.
+            function (Voucher $voucher) use ($bundle, &$errors) {
+                // Check the voucher isn't disbursed, because we can't change those.
+                if ($voucher->bundle && $voucher->bundle->disbursed_at !== null) {
+                    // Throw it into an error.
+                    $errors["codes"][] = $voucher->code;
+                } else {
+                    // Change it's bundle
```
Nikomus:
```suggestion
                    // Change its bundle
```

Also the comment above may be better to say "Check if the voucher is disbursed" - the if is true for disbursed vouchers.

PR #231, opened 2018-12-06
HOTFIX; optimising the RAM we fill querying, by avoiding hydrating fu…

charlesstrange2:
…ll models and being specific.

### DIFF
--- a/app/Trader.php
+++ b/app/Trader.php
@@ -4,6 +4,7 @@ namespace App;
 
 use DB;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Database\Eloquent\SoftDeletes;
 
 class Trader extends Model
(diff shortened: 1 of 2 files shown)


app/Trader.php
```diff
@@ -4,6 +4,7 @@
 
 use DB;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\HasMany;
```
katjam: I guess you only need this for the ide?

app/Trader.php
```diff
@@ -72,18 +73,16 @@ public function vouchersConfirmed()
...
+     * @return \Illuminate\Support\Collection
      */
     public function vouchersWithStatus($status = null)
     {
-        if (empty($status)) {
-            // Get all the trader's assigned vouchers.
-            // Use values() to reset array keys after sort.
-            $vouchers = $this->vouchers->sortByDesc('updated_at')->values()->all();
-        } else {
+        /** @var HasMany $q */
```
katjam: Causing an unused import?

PR #235, opened 2018-12-12
1.5/1212 payment create a place to persist a UUID

charlesstrange2:


app/StateToken.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * The attributes that are mass assignable.
+     *
+     * @var array
+     */
+    protected $fillable = [
+        'uuid'
+    ];
+
+    /**
```
katjam: Remove the boiler plate?

PR #236, opened 2018-12-14
Docker readme update

karalan:


Docker.md
```diff
@@ -5,6 +5,21 @@
...
 
     apt intall docker.io docker-compose
+    
+If you are getting an error to do with the docker composer version, you may need to try another way
+    
+If you dont already have Docker, get it from the docker repo:
+    
+    https://docs.docker.com/install/linux/docker-ce/ubuntu/#install-using-the-repository
+    
+If you are on elementaryOS you need to replace ```$(lsb_release -cs) \``` in step 4 with ```xenial \```
```
charlesstrange2: Does `lsb_release` not work on elementaryOS!? 

I though that was a standard package.

https://refspecs.linuxfoundation.org/LSB_3.0.0/LSB-PDA/LSB-PDA/lsbrelease.html

charlesstrange2: Or does it exist, but not work in the shell that way (aka, is it using bash as a shell, or something more... exotic)?

PR #237, opened 2018-12-17
1.5/1211 payment the payment request submission process generates a persistent UUID token

charlesstrange2:


app/Http/Controllers/API/VoucherController.php
```diff
@@ -84,7 +80,17 @@ function($dirtycode) {
         $failed_rejects = [];
         $vouchers_for_payment = [];
 
+        if ($transition == 'confirm') {
+            // Tidy the StateTokens
+            StateToken::tidy();
+
+            // We'll need a StateToken for Later
+            $stateToken = new StateToken();
```
katjam: Hey - did you ever figure out the tinker crash problem?

app/Http/Controllers/API/VoucherController.php
```diff
@@ -84,7 +80,17 @@ function($dirtycode) {
...
+            $stateToken->save();
+        }
+
         /** @var Voucher $voucher */
+        // TODO: Unsquirrel this ginormo-function.
```
katjam: Don't get the ref - but I am all for simplifying functions.

app/Http/Controllers/API/VoucherController.php
```diff
@@ -106,6 +112,7 @@ function($dirtycode) {
                 // this saves the model too.
                 $voucher->applyTransition($transition);
 
+
```
katjam: Do you want this extra line?

app/Http/Controllers/API/VoucherController.php
```diff
@@ -114,8 +121,15 @@ function($dirtycode) {
                 }
                 // If this is a confirm transition - add to a list
                 // for sending to ARC admin. This is a request for payment.
-                if ($transition === 'confirm') {
+                if ($transition === 'confirm' && $stateToken) {
```
katjam: Do we care if the $stateToken is of type StateToken? Otherwise what is the purpose of the check? Would it be possible to create the token here instead and not need the extra `if === confirm` clause ^^

app/Http/Controllers/API/VoucherController.php
```diff
@@ -134,18 +148,11 @@ function($dirtycode) {
 
         // If there are any confirmed ones... trigger the email.
         if (sizeof($vouchers_for_payment) > 0) {
-            // Todo We need to change something...
-            // If the email fails for some reason,
-            // the User will not be aware.
-            // The Vouchers will be transitioned but the ARC admin won't know.
+            // This email*could* fail; However, ARC admin will eventually be able to see a list of voucher states.
```
katjam: Agree

app/StateToken.php
```diff
@@ -24,6 +27,73 @@ class StateToken extends Model
...
+        parent::__construct($attributes);
+
+        // if it exists and is used
+        if (array_key_exists('uuid', $attributes)) {
+            if (self::isUsedToken($attributes['uuid'])) {
+                $this->uuid = self::generateUnusedToken();
+            }
+        } else {
+            // If it doesn't exist, make one too.
+            $this->uuid = self::generateUnusedToken();
```
katjam: If it's always the same outcome - do we need the branching logic? Maybe just comments instead?

app/StateToken.php
```diff
@@ -24,6 +27,73 @@ class StateToken extends Model
...
+    }
+
+    /**
+     * Makes and checks for an unused token
+     *
+     * @return string
+    */
+    public static function generateUnusedToken()
+    {
+        do {
```
katjam: do .. while loops make me nervous. Is there another way?

app/StateToken.php
```diff
@@ -24,6 +27,73 @@ class StateToken extends Model
...
+            $this->uuid = self::generateUnusedToken();
+        }
+    }
+
+    /**
+     * Makes and checks for an unused token
+     *
+     * @return string
+    */
+    public static function generateUnusedToken()
```
katjam: Are we getting too safety conscious here? What's the likelihood that the generated token is already in use? If it is likely - maybe we need a new generation algorithm.

app/StateToken.php
```diff
@@ -24,6 +27,73 @@ class StateToken extends Model
     protected $hidden = [
     ];
 
+    /**
+     * StateToken constructor. Always end up with a UUID
+     * @param array $attributes
+     */
+    public function __construct(array $attributes = [])
```
katjam: I think if you set up the stateToken uuid to be constrained as unique - you'd be able to rely on $token->save() failing when you try to generate it and not need to do some of these extra checks.

app/StateToken.php
```diff
@@ -24,6 +27,73 @@ class StateToken extends Model
...
+        $expireDate = Carbon::today()->subDays($age)->format('Y-m-d H:i:s');
+        return self::where('created_at', '<', $expireDate)->delete();
+    }
+
+    /**
+     * Checks a UUID has been used
+     * @param $candidate
+     * @return bool
+     */
+    public static function isUsedToken($candidate)
```
katjam: This would be free if uuid was constrained to 'unique' but I am probably missing something obvious.

tests/Unit/Controllers/Api/ApiVoucherControllerTest.php
```diff
@@ -0,0 +1,156 @@
...
+        $this->vouchers = factory(Voucher::class, 'requested', 10)->create();
+        $this->vouchers->each(function ($voucher) {
+            $voucher->applyTransition('order');
+            $voucher->applyTransition('print');
+            $voucher->applyTransition('dispatch');
+        });
+    }
+
+    /** @test */
+    public function testItTidiesOldTokensOnConfirmTransitions()
```
katjam: I feel like this test could be 2 or 3 separate tests - but probably not worth the extra work now to split out.

tests/Unit/Controllers/Api/ApiVoucherControllerTest.php
```diff
@@ -0,0 +1,156 @@
...
+        $data["transition"] = 'confirm';
+
+        // Resubmit
+        $this->actingAs($this->user, 'api')
+            ->json('POST', $route, $data)
+            ->assertStatus(200)
+        ;
+
+        // There should be a token for this request
+        $this->assertEquals(1, StateToken::all()->count());
```
katjam: Seems like you are repeating elements already tested in `testItTidiesOldTokensOnConfirmTransitions()`

tests/Unit/Models/StateTokenTest.php
```diff
@@ -0,0 +1,43 @@
...
+        $specifiedToken = new StateToken(['uuid' => $uuid]);
+        $specifiedToken->save();
+
+        // See that saved correctly
+        $specifiedToken->fresh();
+        $this->assertEquals($uuid, $specifiedToken->uuid);
+    }
+
+    /** @test */
+    public function testItCanCheckTheUUIDisUnique()
```
katjam: Do we also want to test trying to force a save with a non-unique UUID?

app/Http/Controllers/API/VoucherController.php
charlesstrange2: Yes! You cant call a static database function belonging to that class from inside the constructor's scope, apparently.

katjam: Static DB functions are bad anyway. :wink:

app/Http/Controllers/API/VoucherController.php
katjam: Messy ifs are fun!

app/StateToken.php
katjam: Fair.

app/Http/Controllers/API/VoucherController.php
charlesstrange2: No, this gets fired for every voucher, I'd want one StateToken per batch

app/StateToken.php
```diff
@@ -24,6 +27,53 @@ class StateToken extends Model
...
+     * @return string
+    */
+    public static function generateUnusedToken()
+    {
+        do {
+            // TODO: Deal with possibility uuid4() may throw an exception of it's own?
+            try {
+                $candidate = Uuid::uuid4()->toString();
+            } catch (\Exception $e) {
+                // Uuid4() throws exceptions, apparently! Log that and die, I guess?
```
RpprRoger: What. uuid4() throws? What's the error?

app/StateToken.php
RpprRoger: I think uuid v4 is random in the sense that the chance of a conflict is 1 in [`2^122, or 5.3x10^36 (5.3 undecillion) `](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)). Neil is being _very_ cautious here :)

app/StateToken.php
charlesstrange2: "\Ramsey\Uuid\Exception\UnsatisfiedDependencyException if `Moontoast\Math\BigNumber` is not present" in the source.

charlesstrange2: I think it's an optional, but reccomeded _large_ int handler.

app/StateToken.php
RpprRoger: Nice use of a do while :D, it's so rare that they're needed. I suppose this could be expressed using recursion too. Not sure how PHP deals with recursion

app/StateToken.php
RpprRoger: Oh, how frequently does that happen? You would have thought the lib handled that dep missing better if it was optional

PR #238, opened 2018-12-17
1.5/1202 improve visibility of warnings

Nikomus:
- Consistent use of newer warning style
- Test updates
- Quick CSS consistency check

https://trello.com/c/iba3qhUS/1202-improve-visibility-of-warning-of-significant-age-change-and-other-notices-e-v

### DIFF
--- a/resources/views/store/manage_vouchers.blade.php
+++ b/resources/views/store/manage_vouchers.blade.php
@@ -33,11 +33,11 @@
                     </div>
                     <div>
                         @foreach( $family->getNoticeReasons() as $notices )
-                    <p class="v-spaced">
-                        Warning: {{ $notices['count'] }} {{ str_plural($notices['entity'], $notices['count']) }}
-                        is {{ $notices['reason'] }}
-                    </p>
-                    @endforeach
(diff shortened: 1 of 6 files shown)


resources/views/store/manage_vouchers.blade.php
```diff
@@ -27,14 +27,18 @@
...
                 </div>
-                <div class="warning">
-                    @foreach( $family->getNoticeReasons() as $notices )
-                    <p class="v-spaced">
-                        <i class="fa fa-exclamation-circle" aria-hidden="true"></i>
-                        Warning: {{ $notices['count'] }} {{ str_plural($notices['entity'], $notices['count']) }}
-                        is {{ $notices['reason'] }}
-                    </p>
-                    @endforeach
+                <div class="alert-message warning">
```
charlesstrange2: I think you're always producing this box, even if there's no message.
Wrap it in an `@if` based on !empty $family->getNoticeReasons()

Nikomus: Thank you - was fixed in edit page but not here 😞

PR #241, opened 2018-12-19
1.5/1214 payment amend admins payment email to contain a link to the token

charlesstrange2:


charlesstrange2:
StateToken is a UUID attached to a specific VoucherState objects on a Voucher. 

Voucher-<VoucherState>-StateToken

By this mechanism it forms a "handle" we can use to refer to a batch of Vouchers. If we throw that in a payment link, we can click through to a page with a button to transition those on to "reimbursed".

PR #242, opened 2018-12-19
Create route and controller stubs, Test Access control to FM Users

charlesstrange2:
minimum required code for other cards to build on.

### DIFF
--- a/app/Http/Controllers/Store/VoucherController.php
+++ b/app/Http/Controllers/Store/VoucherController.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace App\Http\Controllers\Store;
+
+use App\Http\Controllers\Controller;
+
+class VoucherController extends Controller
+{
+    // This belongs here becuase it's largely about arranging vouchers
+    public function exportMasterVoucherLog()
+    {
+        return response('', 200);
+    }
+}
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Store/VoucherController.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace App\Http\Controllers\Store;
+
+use App\Http\Controllers\Controller;
+
+class VoucherController extends Controller
```
Nikomus: We already have a VoucherController.php in `Controllers/Service/` will this be confusing?

charlesstrange2: True, we do have one there. This is a consequence of smooshing two systems together that functionally operate off the same entity models and routing needs. I knew it as two separated applications, so I guess my model of the codebase still has invisible lines of demarkation in it. 

However, the namespacing works thus far :crossed_fingers: . The only place I think it falls down is in the tests, where phpunits namespacing might need some thought. I had to name something ApiVoucherControlerTest because there was already one VoucherControllerTest one that phpunit couldn't distinguish between.

So, something to watch, and also interesting that a person who joined after might see the app layout in a different light than me

PR #243, opened 2018-12-20
new test for populated bundle

karalan:
note: had to build models separately and pass attributes as using Registration from the Model Factory returns sub Models (Family, Carer) with attributes as strings which then breaks the type checking.

Carded this as a separate issue for investigation as advised by @charlesstrange2

### DIFF
--- a/tests/Unit/Models/BundleModelTest.php
+++ b/tests/Unit/Models/BundleModelTest.php
@@ -9,7 +9,6 @@ use Auth;
 use App\Bundle;
 use App\Registration;
 use Carbon\Carbon;
-use ClassesWithParents\G;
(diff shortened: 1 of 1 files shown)


tests/Unit/Models/BundleModelTest.php
```diff
@@ -3,10 +3,13 @@
 namespace Tests;
 
 
+use App\Carer;
+use App\Family;
 use Auth;
 use App\Bundle;
 use App\Registration;
 use Carbon\Carbon;
+use ClassesWithParents\G;
```
charlesstrange2: Don't think this is required. will cut.

PR #244, opened 2018-12-20
1.5/1216 payment link page

Nikomus:
- This is lacking tests, but otherwise ready for some feedback

https://trello.com/c/52KG8VMy/1216-payment-create-a-view-for-the-payment-link-page


public/store/css/main.css
```diff
@@ -933,6 +933,49 @@ HISTORY PAGE
...
+}
+
+.payment td {
+	font-size: 1.2rem;
+	text-align: center;
+}
+
+.payment button {
+	margin: auto;
+    padding: 0 2rem;
```
Nikomus: 🤔

app/Http/Controllers/Store/PaymentController.php
```diff
@@ -23,8 +25,44 @@ public function __construct()
...
+        // Find the StateToken of a given uuid
+        $state_token = StateToken::where('uuid', $paymentUuid)->first();
+        if ($state_token !== null) {
+
+            // Get the VoucherStates with this StateToken
+            $voucher_states = $state_token
+                ->voucherStates()
+                ->get();
+
+            // Get the voucher codes of states TODO - better
```
charlesstrange2: Solid method for getting the vouchers; Two queries and a bit of collection management :+1:

I think (mildly untested) The "better" solution would be, given `StateToken --< VoucherState >-- Voucher` : 
Add the following function to StateToken, which allows it to reach Vouchers through VoucherStates, like a many-many relationship.

`
   public function vouchers()
    {
        return $this->belongsToMany(Voucher::class, (new VoucherState)->getTable(), 'state_token_id', 'voucher_id');
    }
`
Then do `$vouchers = $state_token->vouchers`

PR #250, opened 2019-01-08
1.5/1225 mvl improve controller stub

charlesstrange2:
Problems overcome
- Use ZipStream instead of ZipArchive to stream zip files to memory
- Don't flush anything to output buffers, or PHPUnit gets very sad. There are workarounds with PHPUnits directives *if* your output doesn't use binary files (with `null` characters in).


app/Http/Controllers/Store/VoucherController.php
```diff
@@ -3,12 +3,133 @@
...
-    // This belongs here becuase it's largely about arranging vouchers
+    // This belongs here because it's largely about arranging vouchers
     public function exportMasterVoucherLog()
     {
-        return response('', 200);
+        $dashboard_route = URL::route('store.dashboard');
+
+        // Setup the Storage dir
+        $disk = Storage::disk(config('arc.mvl_disk'));
+        $archiveName = config('arc.mvl_filename');
```
katjam: Only nit picking... mixing camel case with underscore for var names in the same file - hinders reading a tiny bit.

app/Http/Controllers/Store/VoucherController.php
```diff
@@ -3,12 +3,133 @@
...
+                    $message = ('not a zip archive');
+                    break;
+                case ZipArchive::ER_INCONS:
+                    $message = ('consistency check failed');
+                    break;
+                case ZipArchive::ER_CRC:
+                    $message = ('checksum failed');
+                    break;
+                default:
+                    $message = ('error ' . $check);
```
katjam: will check always be a string?

app/Http/Controllers/Store/VoucherController.php
```diff
@@ -3,12 +3,133 @@
...
+            }
+
+            // Log that error
+            Log::error(
+                // Leading `.` as url() gives a leading `/` and the storage is not off root
+                "ZipArchive cannot open archive at '." .
+                // Using url() to prevent revealing the disk tree
+                $disk->url('$archiveName') .
```
katjam: Should the quotes be around that var?

app/Http/Controllers/Store/VoucherController.php
```diff
@@ -3,12 +3,133 @@
...
+            // go back to dashboard with a
+            return redirect($dashboard_route)
+                ->with('error_message', "Sorry, the export file was unreadable. Please contact support.");
+        }
+
+        $unpackedFiles = [];
+
+        // Iterate through files in the zip.
+        // TODO : Change to $zip->count() if we move to PHP 7.2
+        for ($i = 0; $i < $storedZip->numFiles; $i++) {
```
katjam: By the time we get here we've exited if we don't have a valid $storedZip, yes?

app/Http/Controllers/Store/VoucherController.php
```diff
@@ -3,12 +3,133 @@
...
+            $sourceName = $storedZip->getNameIndex($i);
+
+            // Open the stream for that file
+            $fileStream = $storedZip->getStream($sourceName);
+            if (!$fileStream) {
+                Log::error("ZipArchive cannot read compressed item '" . $sourceName . "'' by stream");
+                continue;
+            }
+
+            // Get the file contents to memory. Bit of a farce but i don't want to extract to a temp file
```
katjam: I agree this is a bit too granular faff - but works and fine.

katjam: Seems like there should be a higher level utility for this available.

tests/Unit/Routes/StoreRoutesTest.php
```diff
@@ -483,8 +484,12 @@ public function testMVLRouteGuard()
 
         // See page permit FM User to get from the dashboard.
         $this->actingAs($fmuser, 'store')
-            ->visit(URL::route('store.dashboard'))
+            ->visit($dashboard_route)
             ->get($route)
+            // It's going to 302->200 if there's a problem finding/reading the file (which isn't there).
+            // However, we need to know that it *can* hit the right route
+            // Tests for *what gets returned are elsewhere.
+            ->followRedirects()
```
katjam: :+1:

PR #261, opened 2019-07-21
Pre-mangle voucher data to clean spaces before we run the validations.

charlesstrange2:
https://trello.com/c/P8hFB5ET/1245-remimplement-hotfix-patch-vouchers-with-spaces-in-the-code-are-confusing-the-system-vvv-ee

### DIFF
--- a/app/Http/Requests/StoreAppendBundleRequest.php
+++ b/app/Http/Requests/StoreAppendBundleRequest.php
@@ -2,8 +2,8 @@
 
 namespace App\Http\Requests;
 
-use Illuminate\Foundation\Http\FormRequest;
 use App\Voucher;
+use Illuminate\Foundation\Http\FormRequest;
 
 class StoreAppendBundleRequest extends FormRequest
 {
--- a/app/Http/Requests/StoreAppendBundleRequest.php
+++ b/app/Http/Requests/StoreAppendBundleRequest.php
@@ -3,6 +3,7 @@
 namespace App\Http\Requests;
 
 use Illuminate\Foundation\Http\FormRequest;
+use App\Voucher;
 
 class StoreAppendBundleRequest extends FormRequest
 {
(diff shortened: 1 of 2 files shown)


app/Http/Requests/StoreAppendBundleRequest.php
```diff
@@ -3,6 +3,7 @@
 namespace App\Http\Requests;
 
 use Illuminate\Foundation\Http\FormRequest;
+use App\Voucher;
```
katjam: :smile: You know me - I like these to be ABC - but not stopping you. Hardly relevant with 2 imports!

app/Http/Requests/StoreAppendBundleRequest.php
```diff
@@ -37,4 +38,21 @@ public function rules()
...
         return $rules;
     }
+
+    protected function prepareForValidation()
+    {
+        // get the input and remove null/empty values.
+        // TODO: 5.5 upgrade will need to change this to 'all' with params.
+        $input = array_filter(
+            $this->only(['start', 'end']),
+            'strlen'
```
katjam: Not too keen on 0 == false but I think it's OK here.

tests/Unit/Controllers/Store/BundleControllerTest.php
```diff
@@ -397,4 +397,32 @@ public function testICannotDisburseAnEmptyBundle()
             ->seePageIs($route)
             ->assertResponseStatus(200);
     }
+
+    /** @test */
+    public function itCanAcceptAndCleanVouchersWithSpacesIn()
```
katjam: Do you also want to test that `null` and `""` values will get cleared from the list?

app/Http/Requests/StoreAppendBundleRequest.php
charlesstrange2: it's apparently a well known array_filter trick :man_shrugging:

tests/Unit/Controllers/Store/BundleControllerTest.php
```diff
@@ -397,4 +397,79 @@ public function testICannotDisburseAnEmptyBundle()
...
+        /** @var Bundle $currentBundle */
+        // Get our currentBundle
+        $currentBundle = $this->registration->currentBundle();
+
+        // See that it's got one voucher.
+        $this->assertEquals(1, $currentBundle->vouchers()->count());
+    }
+
+    /** @test */
+    public function itHasSparseFormDataCleanedBeforeProcessing()
```
katjam: Title is a bit opaque - but otherwise. :+1:

PR #267, opened 2019-07-24
1306 view children centres

alexalala:
https://trello.com/c/CEZAMsE4/1306-view-children-centres

🐸


routes/service.php
```diff
@@ -40,6 +40,10 @@
         'as' =>'admin.vouchers.create',
         'uses' => 'Admin\VouchersController@create',
     ]);
+    Route::get('centres', [
+        'as' =>'admin.centres.centres_view',
```
charlesstrange2: would be better as `admin.centres.index`, like the one for vouchers.

PR #269, opened 2019-07-25
1.6/1253 add a limit on voucher adds

charlesstrange2:
https://trello.com/c/yJQ6TD8I/1253-add-a-limit-on-how-many-vouchers-can-be-added-in-one-go-vvv-ee

### DIFF
--- a/app/Voucher.php
+++ b/app/Voucher.php
@@ -81,6 +81,45 @@ class Voucher extends Model
     }
 
     /**
+     * Will generate a code range from start to end in the start range
+     *
+     * @param string $start voucher code to start with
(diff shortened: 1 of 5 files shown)


app/Voucher.php
```diff
@@ -80,6 +80,45 @@ function ($code) {
...
     }
 
+    /**
+     * Will generate a code range from start to end in the start range
+     *
+     * @param string $start voucher code to start with
+     * @param string $end voucher code to end with
+     * @return array
+     */
+    public static function generateCodeRange($start, $end = "")
```
katjam: Much better here.

PR #272, opened 2019-07-26
View sponsors page created and routed

alexalala:
https://trello.com/c/Wb0DGkwm/1294-view-sponsors

🐣

### DIFF
--- a/resources/views/service/sponsors/index.blade.php
+++ b/resources/views/service/sponsors/index.blade.php
@@ -11,7 +11,7 @@
             <thead>
                 <tr>
                     <th>Name</th>
-                    <th>RVID Prefix</th>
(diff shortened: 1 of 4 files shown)


resources/views/service/sponsors/index.blade.php
```diff
@@ -0,0 +1,30 @@
...
+
+    @include('service.includes.sidebar')
+
+    <div id="main-content">
+        <h1>Sponsors</h1>
+        <table class="table table-striped">
+            <thead>
+                <tr>
+                    <th>Name</th>
+                    <th>RVID Prefix</th>
```
charlesstrange2: This isn't the `RVID Prefix`, that's for family registrations. This should probably be `Voucher Prefix`

PR #274, opened 2019-07-26
1.6/1230 exp laravel54 55lts upgrade

charlesstrange2:
https://trello.com/c/hY5drUJQ/1230-laravel-54-55lts-upgrade-vv-eee

This is a biggy,

After merge we'll have to issue
`php artisan view:clear`
`php artisan cache:clear`
and remove contents of ./storage/framework/sessions, 

Further, using node 8.x LTS (or , if you have it, `nvm install lts/carbon && nvm use`)
`rm -f node_modules`
`yarn install`
`npm run prod` (possibly, twice as it may abort and compile the local sass binaries the firt time, instead)


tobybatch:
Tests all pass in dev on 7.2 but when testing in the vagrant...

```
$ composer install
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Nothing to install or update
Package phpoffice/phpexcel is abandoned, you should avoid using it. Use phpoffice/phpspreadsheet instead.
Package phpunit/phpunit-mock-objects is abandoned, you should avoid using it. No replacement was suggested.
Generating optimized autoload files
Carbon 1 is deprecated, see how to migrate to Carbon 2.
https://carbon.nesbot.com/docs/#api-carbon-2
    You can run './vendor/bin/upgrade-carbon' to get help in updating carbon and other frameworks and libraries that depend on it.
> Illuminate\Foundation\ComposerScripts::postAutoloadDump
> @php artisan package:discover

In DuskServiceProvider.php line 43:
                                           
  It is unsafe to run Dusk in production.  
                                           

Script @php artisan package:discover handling the post-autoload-dump event returned with error code 1
```

This error is just luck.  The ansible call composer install before setting up the .env file so the app defaults to prod settings.

README.md
```diff
@@ -2,13 +2,11 @@
 ## About ARC Voucher Service/API
 ARCV Service is the service portal and API for ARCV Market.
 
-[![Build Status](https://travis-ci.org/neontribe/ARCVMarket.svg?branch=0.2/release)](https://travis-ci.org/neontribe/ARCVMarket.svg?branch=master)
```
katjam: Why remove?

app/Console/Kernel.php
```diff
@@ -7,19 +7,6 @@
 
 class Kernel extends ConsoleKernel
 {
-    /**
-     * The Artisan commands provided by your application.
-     *
-     * @var array
-     */
-    protected $commands = [
```
katjam: Did you intend to move these to the `routes/console.php`? Or deleted altogether?

app/Http/Controllers/Store/RegistrationController.php
```diff
@@ -424,8 +424,7 @@ function ($child) {
 
         // Expect only filled fm variables
         $fm = array_filter(
-            $request->only('fm_privacy'),
-            // Previously: $request->only('fm_chart', 'fm_diary', 'fm_privacy'),
+            $request->all('fm_privacy'),
```
katjam: Does `all()` take string or this needs to be `[]`?

resources/assets/js/app.js
```diff
@@ -6,17 +6,3 @@
  */
 
 require('./bootstrap');
-
```
katjam: Assuming this is how you want it - and don't need Vue.

tests/Unit/Routes/StoreRoutesTest.php
```diff
@@ -390,10 +390,13 @@ public function testSessionUpdateGate()
             ->assertResponseStatus(200);
     }
 
-
     /** @test */
     public function testRegistrationFamilyUpdateGate()
     {
+        // TODO complete or discard
```
katjam: Assuming you want this to go in as a todo

README.md
charlesstrange2: becuase it relates to the market app, and this is the service/store repo

app/Console/Kernel.php
charlesstrange2: I removed it because the auto-discovery is in play AND i didn't want to null the array that discovery makes.

app/Http/Controllers/Store/RegistrationController.php
charlesstrange2: it's cheeky and takes _either_ - there's some code in all() that works out if you asked for a single string attribute or put a bunch in an array.

resources/assets/js/app.js
charlesstrange2: Nope, we don't use it.

tests/Unit/Routes/StoreRoutesTest.php
charlesstrange2: yeah, it's on a list for tests-to-complete. :-(

README.md
katjam: Fair!

app/Console/Kernel.php
katjam: Sounds good.

app/Http/Controllers/Store/RegistrationController.php
katjam: I think `only` did that too. Just checking since somewhere else we use an array with a single item.

charlesstrange2: only changed in 5.5 to NOT return an empty property IF the property was not found.
To maintain 5.4 only('blah') behaviour, you have ti use all('blah').

PR #291, opened 2019-08-14
1.6/1316 bug voucher add form times out becuase controller takes too long

charlesstrange2:
Tidies up the validation
Adds a form requests
Massively speeds up voucher inserts and state creation by batch inserting.
This is valuable when adding 100,000s vouchers in one go (which has happened recently)

https://trello.com/c/doNOciRC/1316-bug-voucher-add-form-times-out-becuase-controller-takes-too-long


app/Http/Controllers/Service/Admin/VouchersController.php
```diff
@@ -45,69 +41,109 @@ public function create()
...
-
-        Validator::make($request->all(), $voucher_rules, $messages)->validate();
-
-        $codes = range($request['start'], $request['end']);
-
-        $sponsor = Sponsor::find($request['sponsor_id'])->first();
-        $shortcode = $sponsor->shortcode;
+        // Setup some variables
+        $input = $request->all();
+        $user_id = auth()->id();
```
katjam: Do you want an abort if no user?

app/Http/Controllers/Service/Admin/VouchersController.php
```diff
@@ -45,69 +41,109 @@ public function create()
...
         $now_time = Carbon::now();
-        foreach ($codes as $c) {
-            $v = new Voucher();
-            $v->code = $shortcode . $c;
-            $v->sponsor_id = $request['sponsor_id'];
-            $v->currentstate = 'requested';
-            $v->created_at = $now_time;
-            $v->updated_at = $now_time;
-            $new_vouchers[] = $v->attributesToArray();
+        $maxStep = 1000;
```
katjam: Neat!

app/Http/Controllers/Service/Admin/VouchersController.php
```diff
@@ -45,69 +41,109 @@ public function create()
...
+                $v->sponsor_id = $sponsor->id;
+                // Set straight to printed; we're faking the process for speed.
+                $v->currentstate = 'printed';
+                $v->created_at = $now_time;
+                $v->updated_at = $now_time;
+                $new_vouchers[] = $v->attributesToArray();
+                unset($v);
+            }
+            // Batch insert raw vouchers.
+            // Todo : there's NO "this voucher already exists" checking!!
```
katjam: How likely is it that this will be an issue?

app/Http/Controllers/Service/Admin/VouchersController.php
charlesstrange2: in this case auth()->id being null isn't a showstopper, also, should always be auth'd by this point

app/Http/Controllers/Service/Admin/VouchersController.php
charlesstrange2: Ugh, could be a serious problem, BUT i think we'll fix it in the next 24 hours OR in the next dev sprint. ARC have promised to be careful.

katjam: :grimacing: It'll be ok.

PR #293, opened 2019-08-15
1.6/1336 bug content of bad code problem error message

charlesstrange2:
https://trello.com/c/jvgUIsXi/1336-bug-content-of-bad-code-problem-error-message


charlesstrange2:
The problem here was that stealing vouchers from disbursed bundles didn't have it's own error and was using the error for "invalid code"

PR #297, opened 2019-08-28
1.7/1303 revisit streaming/decrypting the MVL (V/EE)

Hoolean:
# The Problem

The system finds generating the Master Voucher Log taxing. Because of this, it is instructed to generate the log in the dead of night, to avoid any significant service interruption.

As storage and security dictate that we should not leave the log sitting on the disk, some time ago, the decision was made to compress and encrypt it. As the log has grown in size, post-processing the log on-demand to serve to the user has become a performance issue in itself.

Optimising the initial log generation or rejigging our production server's security policy would be favourable solutions, however, short-term improvements will go far until sufficient resources are available.

# Progress in Resource Usage

## 100% of Original Memory Usage

Memory is the main expense when the system serves the logs, as they can never be stored in plaintext on the filesystem (even temporarily). The entire contents of the logs are stored in memory a few times over, while the system decompresses, decrypts and recompresses them as a whole, which is what has become infeasible.

## 75% of Original Memory Usage (tested)

With a streaming approach, memory use is constant relative to the size of the log: only a small working buffer needs space at any given time.

```
[ Decompress ] -> [ Decrypt ] -> [ Recompress ]
```

Reimplementing any part of the process to use streaming prevents the file from being stored in memory at that stage. If every part of the process could use streaming, we could achieve full scalability, only storing working buffers at any given time.

Without any significant architectural change, we can stream the final ZIP to the user, preventing it from being stored in memory. This is what is implemented in this pull request so far.

*NOTE: The file size is no longer known in advance to be sent to the user, breaking download progress bars. Is this an acceptable cost?*

## Going Constant

Zipping and unzipping are highly amenable to a streaming approach. Unfortunately, the Laravel library that we are using for encryption and decryption is not.

If we moved to an encryption library that offered streams as a primitive:
* Every part of the process would be scalable
* It would be performant to compress logs prior to encryption, instead of after, as we do now
  * The decompression and recompression steps could be removed from serving the logs entirely
  * The download progress bar would be restored
* The nightly creation of logs would become amenable to streaming too, opening up many optimisation routes and paving the way for friendlier caching techniques
 
### [libsodium](https://download.libsodium.org/doc/)

`libsodium` appears to be a practical and dependable cryptography library and would not require too much implementation work to use.

#### Disadvantages of Library

* The simple `encrypt()` and `decrypt()` functions we use now would have to be replaced with stream-friendly code
* The library would operate outside of Laravel's build-in cryptography
* CPU usage may be heavier if hardware decryption was used before
* Complexity makes cryptography easier to misuse

#### Advantages of Library

* [Popular and well-supported](https://github.com/jedisct1/libsodium)
* [Encrypted streams as a feature](https://libsodium.gitbook.io/doc/secret-key_cryptography/secretstream)
* API is not too complicated for our use-case

### DIFF
--- a/app/Http/Controllers/Store/VoucherController.php
+++ b/app/Http/Controllers/Store/VoucherController.php
@@ -6,6 +6,7 @@ use App\Http\Controllers\Controller;
 use Carbon\Carbon;
 use Illuminate\Support\Facades\Storage;
 use Log;
+use Symfony\Component\HttpFoundation\StreamedResponse;
 use URL;
 use ZipArchive;
 use ZipStream\ZipStream;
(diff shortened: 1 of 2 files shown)


charlesstrange2:
@Hoolean - excellent documentation - thankyou :+1:

PR #300, opened 2019-09-02
1.7/1303 revisit streaming/decrypting the MVL (V/EE) 

Hoolean:
# Progress

This PR implements the improvements suggested in #297, serving the Master Voucher Log with a streaming approach at every stage. As a result, memory usage rests comfortably at only a few megabytes and is now constant for any log size.

## New Flow

All spreadsheets are now compressed prior to encryption, as opposed to vice-versa.

Encryption and decryption occurs on output and input streams respectively, avoiding the need for either of these to hold the entire file contents in memory at any given time. This is achieved with:
* A stream wrapper for the log generation process
* Direct usage of the decryption streaming primitive in the log serving process.

### Log Creation

```
Database -> Spreadsheet -> ZIP Stream -> Secret Stream Wrapper -> File on Disk
```

### Log Serving

```
File on Disk -> Decryption Stream -> HTTP Response Stream
```

## Caveats
* The log's file size is fiddly to calculate prior to decryption, resulting in a download progress bar that lacks an estimation of progress and time remaining.
* A custom error page cannot be shown after the system has started serving the log, as an ongoing HTTP streaming response cannot be modified very semantically.
* The encryption library is not integrated into Laravel (but is part of the PHP standard library, as of PHP 7.2).

## Features
* The approach is scalable and should not require future optimisation.
* The log file is now compressed to a far smaller size, prior to encryption, reducing its use of memory, bandwidth and storage resources.
* The implementation of streaming encryption is highly reusable and could help reduce memory usage in other processes (including the generation of the file).

### DIFF
--- a/app/Wrappers/SecretStreamWrapper.php
+++ b/app/Wrappers/SecretStreamWrapper.php
@@ -1,0 +1,187 @@
+<?php
+
+
+namespace App\Wrappers;
+
+use Illuminate\Support\Str;
+
+/**
+ * This stream wrapper allows the creation of a resource that will encrypt all data forwarded to another resource, such
+ * as a file or socket.
+ *
+ * Stream wrappers are fairly poorly documented in the PHP manual. They can be thought of as abstract files, which may
+ * or may not wrap another resource.
+ *
+ * https://www.php.net/manual/en/class.streamwrapper.php - for the "class" this one extends
+ * https://www.php.net/manual/en/stream.streamwrapper.example-1.php - for an example of another
+ * https://www.php.net/manual/en/stream.streamwrapper.example-1.php#99283 - for a more straight forward example of another
+ *
+ * The encryption used is provided by `libsodium`, a popular and robust library that can have its bindings compiled into
+ * PHP itself (as of version 7.2). The particular primitive is secret stream, which allows messages of known size from a
+ * larger source to be decrypted individually, with guarantees of order and completeness. This is important if we do not
+ * want the hold the entirety of a file in memory at any given time.
+ *
+ * Documentation is similarly poor for secret streams in PHP but examples are common for other languages. They can be
+ * thought of as encryption and decryption with state.
+ *
+ * https://libsodium.gitbook.io/doc/secret-key_cryptography/secretstream - description of the primitive and C examples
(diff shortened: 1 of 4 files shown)


app/Wrappers/SecretStreamWrapper.php
```diff
@@ -0,0 +1,187 @@
...
+    public static function getKey() {
+        $key = config('app.key');
+
+        // Sometimes, the key is encoded in base64. Decrypt it if this is the case.
+        if (Str::startsWith($key, 'base64:')) {
+            $key = base64_decode(substr($key, 7));
+        }
+
+        // Check that the key is the write length.
+        // TODO : Can we guarantee that app.key will be configured as 32 random bytes?
```
charlesstrange2: If people use `artisan key:generate` then yes, it should be.

app/Wrappers/SecretStreamWrapper.php
```diff
@@ -0,0 +1,187 @@
...
+
+        // Flush out the final message, tagged as such.
+        if (strlen($this->buffer) > 0) {
+            $encrypted = sodium_crypto_secretstream_xchacha20poly1305_push(
+                $this->stream,
+                $this->buffer,
+                '', // This is an inactive placeholder.
+                SODIUM_CRYPTO_SECRETSTREAM_XCHACHA20POLY1305_TAG_FINAL
+            );
+            fwrite($this->wrapped, $encrypted); // TODO : Can an error here be caught somehow?
```
charlesstrange2: frwite, being a fairly basic function returns falsey (false or, possibly 0) as an error state. not terribly useful

Hoolean: From testing, after writing the comment, it seemed that Laravel was eager to treat these errors pretty seriously without having to spot them manually. The error would be spotted by the single user on serving the file as well, so I think I'll leave this uncaught :+1:

app/Wrappers/SecretStreamWrapper.php
Hoolean: I don't see much of a risk of anything insecure but also the right length being used here. We'll get a nice error spat out if the length is wrong too, so I think we should be able to leave this as is :+1:

PR #326, opened 2019-09-18
1.7/1347 data adding sent vouchers

charlesstrange2:
Adds the store function, validation form request and voucher code to
- take in start and end voucher number, a centre and a dispatch date
- turn that into a range of vouchers
- check it against the available vouchers for delivery
- fake a voucher transition for printed->dispatched with a batch insert voucherstates for the voucher 
- create a delivery record
- throw errors back at the user when it breaks,

Automated tests hard to write, because the means of quickly checking if a voucher range is available for delivery uses mysql variables. I had to abaondon eloquent becuase active record isn't up to manipulating 1,000,000 records in one go.


charlesstrange2:
Damn! forgot some error reporting.

PR #327, opened 2019-09-18
1.7/1347 data adding sent vouchers second try

charlesstrange2:
Adds the store function, validation form request and voucher code to

- take in start and end voucher number, a centre and a dispatch date
- turn that into a range of vouchers
- check it against the available vouchers for delivery
- fake a voucher transition for printed->dispatched with a batch insert voucherstates for the voucher 
- create a delivery record
- throw errors back at the user when it breaks,

Automated tests hard to write, because the means of quickly checking if a voucher range is available for delivery uses MySQL variables. I had to abandon eloquent because active record isn't up to manipulating 1,000,000 records in one go.


app/Http/Controllers/Service/Admin/DeliveriesController.php
```diff
@@ -2,32 +2,73 @@
...
+     * @param $shortcode
+     * @return bool
+     * @throws Throwable
+     */
+    public static function rangeIsUndelivered($start, $end, $shortcode)
+    {
+        $undeliveredRanges = Voucher::getUndeliveredVoucherRangesByShortCode($shortcode);
+
+        // If NOT an empty array
+        if (!empty($undeliveredRanges)) {
```
Hoolean: I think this condition may be redundant

app/Http/Controllers/Service/Admin/DeliveriesController.php
```diff
@@ -36,8 +77,110 @@ public function create()
...
+                $now_time = $delivery->created_at;
+                $user_id = auth()->id();
+                $user_type = class_basename(auth()->user());
+
+                // Bulk update VoucherStates for speed.
+                Voucher::select('id')
+                    ->where('currentState', 'printed')
+                    ->whereNull('delivery_id')
+                    ->whereBetween(
+                        DB::raw("cast(replace(code, '{$start["shortcode"]}', '') as signed)"),
```
Hoolean: This allows for injection with an input similar to: `SELECT/**/*/**/FROM/**/users12356789`

app/Http/Controllers/Service/Admin/DeliveriesController.php
```diff
@@ -36,8 +77,110 @@ public function create()
...
+                            // Insert this batch of vouchers.
+                            VoucherState::insert($states);
+                        }
+                    );
+
+                // Get all the vouchers in the range and update them with the delivery Id and state
+                Voucher::where('currentState', 'printed')
+                    ->whereNull('delivery_id')
+                    ->whereBetween(
+                        DB::raw("cast(replace(code, '{$start["shortcode"]}', '') as signed)"),
```
Hoolean: Same issue as above

app/Voucher.php
```diff
@@ -142,6 +146,99 @@ public static function splitShortcodeNumeric(string $code)
...
+                        SELECT
+                            @initial_serial := if(serial - @previous = 1, @initial_serial, serial) as initial_serial,
+                            @initial_id := if(serial - @previous = 1, @initial_id, id) as initial_id,
+                            @previous := serial as serial,
+                            id as final_id
+                        FROM (
+                        
+                            SELECT id, cast(replace(code, '{$shortcode}', '') as signed) as serial
+                            FROM vouchers
+                            WHERE code REGEXP '^{$shortcode}[0-9]+\$'
```
Hoolean: Same issue as above

app/Voucher.php
```diff
@@ -142,6 +146,99 @@ public static function splitShortcodeNumeric(string $code)
...
+                    FROM (
+                        
+                        SELECT
+                            @initial_serial := if(serial - @previous = 1, @initial_serial, serial) as initial_serial,
+                            @initial_id := if(serial - @previous = 1, @initial_id, id) as initial_id,
+                            @previous := serial as serial,
+                            id as final_id
+                        FROM (
+                        
+                            SELECT id, cast(replace(code, '{$shortcode}', '') as signed) as serial
```
Hoolean: Same issue as above

app/Http/Controllers/Service/Admin/DeliveriesController.php
charlesstrange2: True! it returns an array, which may be empty, so... we wouldn't iterate over it.

app/Http/Controllers/Service/Admin/DeliveriesController.php
charlesstrange2: The voucher the shortcode is split from is validated as real before the controller function is hit. If it doesn't exist it's going to get caught earlier than this.

app/Http/Controllers/Service/Admin/DeliveriesController.php
charlesstrange2: As above

app/Voucher.php
charlesstrange2: as above

app/Voucher.php
charlesstrange2: As above

PR #329, opened 2019-09-18
1.7/1351 hotfix srtpad the correct number of 0s

charlesstrange2:
This makes it possible for users to add zero padded voucher codes.
It will pad to the length of the end code
It will error apporpriately.

### DIFF
--- a/app/Http/Controllers/Service/Admin/VouchersController.php
+++ b/app/Http/Controllers/Service/Admin/VouchersController.php
@@ -55,10 +55,9 @@ class VouchersController extends Controller
         $maxStep = 1000;
         $start = $input['start-serial'];
         $end = $input['end-serial'];
-        $pad = strlen($end);
 
         // Calculate the number of codes we need
-        $numCodes = $start - $end;
+        $numCodes = $end - $start;
 
         $step = 1;
         if ($numCodes > 1) {
@@ -76,7 +75,7 @@ class VouchersController extends Controller
         );
 
         // Add the range to the end.
-        if (!in_array($start, $chunks)) {
+        if (!in_array($end, $chunks)) {
             $chunks[] = $end;
         }
 
--- a/app/Http/Controllers/Service/Admin/VouchersController.php
+++ b/app/Http/Controllers/Service/Admin/VouchersController.php
@@ -53,9 +53,12 @@ class VouchersController extends Controller
         $sponsor = Sponsor::findOrFail($input['sponsor_id']);
         $now_time = Carbon::now();
         $maxStep = 1000;
+        $start = $input['start-serial'];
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Service/Admin/VouchersController.php
```diff
@@ -53,9 +53,12 @@ public function storeBatch(AdminNewVoucherRequest $request)
...
         $now_time = Carbon::now();
         $maxStep = 1000;
+        $start = $input['start-serial'];
+        $end = $input['end-serial'];
+        $pad = strlen($end);
 
-        // Calulate the number of codes we need
-        $numCodes = $input['end'] - $input['start'];
+        // Calculate the number of codes we need
+        $numCodes = $start - $end;
```
Hoolean: This looks like it might have been accidentally flipped

app/Http/Controllers/Service/Admin/VouchersController.php
```diff
@@ -67,14 +70,14 @@ public function storeBatch(AdminNewVoucherRequest $request)
...
-            $input['end'],
+            $start,
+            $end,
             $step
         );
 
         // Add the range to the end.
-        if (!in_array($input['end'], $chunks)) {
-            $chunks[] = $input['end'];
+        if (!in_array($start, $chunks)) {
```
Hoolean: Same issue as above

app/Http/Requests/AdminNewVoucherRequest.php
```diff
@@ -27,10 +27,37 @@ public function rules()
...
+        $this->merge($input);
+    }
+
+    // Friendlier messages to reference computed fields to normal ones.
+    public function messages()
+    {
+        return  [
+            'start-serial.between.numeric' => 'The start must be between :min and :max.',
+            'end-serial.between.numeric' => 'The end must be between :min and :max.',
+            'end-serial.ge_field' => 'The end must be greater than or equal to start.'
```
Hoolean: This should probably be 'The end must be greater than or equal to the start' or 'End must be greater than or equal to start'

app/Http/Controllers/Service/Admin/VouchersController.php
charlesstrange2: yep!

app/Http/Controllers/Service/Admin/VouchersController.php
charlesstrange2: yep!

app/Http/Requests/AdminNewVoucherRequest.php
charlesstrange2: Purely copy

PR #332, opened 2019-09-19
1.7/1408 extended rules

charlesstrange2:
Add two rules to detect children being over school and secondary school age
Add them to the rulesset
Add tests for them
Make factories for children that fit that.
Change the language on the school and school age to be more friendly

### DIFF
--- a/app/Services/VoucherEvaluator/Evaluations/ChildIsSchoolAge.php
+++ b/app/Services/VoucherEvaluator/Evaluations/ChildIsSchoolAge.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace App\Services\VoucherEvaluator\Evaluations;
+
+use App\Specifications\IsBorn;
+use App\Specifications\IsUnderStartDate;
+use Carbon\Carbon;
+use Chalcedonyt\Specification\AndSpec;
+use Chalcedonyt\Specification\NotSpec;
+
+class ChildIsSchoolAge extends BaseChildEvaluation
+{
+    const REASON = 'school age (ineligible)';
+    private $specification;
+
+    public function __construct(Carbon $offsetDate = null, $value = 3)
+    {
+        parent::__construct($offsetDate, $value);
+
+        $this->specification = new AndSpec(
+            new IsBorn(),
+            // Child school start date
+            new notSpec(new IsUnderStartDate($this->offsetDate, 5, config('arc.school_month')))
+        );
+    }
+
+    public function test($candidate)
+    {
+        parent::test($candidate);
+
+        return ($this->specification->isSatisfiedBy($candidate))
+            ? $this->success()
+            : $this->fail()
+        ;
+    }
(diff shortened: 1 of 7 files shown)


app/Services/VoucherEvaluator/Evaluations/ChildIsSchoolAge.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    const REASON = 'school age (ineligible)';
+    private $specification;
+
+    public function __construct(Carbon $offsetDate = null, $value = 3)
+    {
+        parent::__construct($offsetDate, $value);
+
+        $this->specification = new AndSpec(
+            new IsBorn(),
```
Hoolean: I think this is redundant - if a child is over school age they're also born - but not doing any harm.

app/Services/VoucherEvaluator/Evaluations/ChildIsSecondarySchoolAge.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    const REASON = 'secondary school age (ineligible)';
+    private $specification;
+
+    public function __construct(Carbon $offsetDate = null, $value = 3)
+    {
+        parent::__construct($offsetDate, $value);
+
+        $this->specification = new AndSpec(
+            new IsBorn(),
```
Hoolean: Same redundancy as above.

tests/Unit/Services/VoucherEvaluator/VoucherEvaluatorTest.php
```diff
@@ -176,6 +197,27 @@ public function itNoticesWhenAChildIsAlmostSchoolAge()
         $this->assertContains(self::NOTICE_TYPES['ChildIsAlmostSchoolAge'], $notices);
     }
 
+    /** @test */
+    public function itNoticesWhenAChildIsAlmostSecondarySchoolAge()
+    {
+        // Need to change the values we use for school start to next month's integer
+        Config::set('arc.school_month', Carbon::now()->addMonth(1)->month);
```
Hoolean: If it is December and this month wraps around to January, this test will not behave as expected.

app/Services/VoucherEvaluator/Evaluations/ChildIsSchoolAge.php
charlesstrange2: In this rule, that's true; we're never going to have a 5 year pregnancy. I'll mark it for removal at some point. In other rules it helps check for overdue pregnancies.

tests/Unit/Services/VoucherEvaluator/VoucherEvaluatorTest.php
charlesstrange2: As discussed, Card and put in tech debt to be dealt with next heartbeat.

PR #338, opened 2019-12-03
1.8/1443 fix credentials in stacktraces

charlesstrange2:
Slightly sad that i cant test exception logs to the file - `Log::shoudlReceive()` can't detect the way the exception handling system invokes it's own logger, rather than the Log interface. This means the mocker can't easily intercept. 

Have tried to track the decision to log full stacks or not with some reflection instead.

https://trello.com/c/zxWFWUfd/1443-fix-for-security-issue-from-14-11-2019-vvv-ee

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -68,13 +68,14 @@ class Handler extends ExceptionHandler
      *
      * @throws Exception
      * @param  Exception  $e
-     * @return void
+     * @return mixed
      */
     public function report(Exception $e)
     {
         if (!$this->shallWeStackTrace($e)) {
             try {
                 // Should build an appropriate logger for our config;
+                // Alas, Log::shouldReceive can't detect this.
                 $logger = $this->container->make(LoggerInterface::class);
             } catch (Exception $ex) {
                 // Whoops, the logger didn't get made, throw the original exception
@@ -90,7 +91,7 @@ class Handler extends ExceptionHandler
                 .$e->getTrace()[0]['file']
             );
         } else {
-            parent::report($e);
+            return parent::report($e);
         }
     }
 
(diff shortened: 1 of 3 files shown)


app/Exceptions/Handler.php
```diff
@@ -18,33 +28,79 @@ class Handler extends ExceptionHandler
...
+        OAuthServerException::class,
+    ];
+
+    /**
+     * Check an exception is worth a stacktrace
+     *
+     * @param Exception $e Exception
+     * @return bool;
+     */
+    protected function shallWeStackTrace(Exception $e)
```
katjam: Cute

tests/Unit/Exceptions/ExceptionHandlerTest.php
```diff
@@ -0,0 +1,76 @@
...
+use Tests\StoreTestCase;
+
+class ExceptionHandlerTest extends StoreTestCase
+{
+
+    private $exceptions = [];
+
+    protected function setUp()
+    {
+        // Tests reach into the container, and get upset when run in bulk with other tests.
```
katjam: :cry:

tests/Unit/Exceptions/ExceptionHandlerTest.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /** @test */
+    public function itShouldStackTraceNotInProduction()
+    {
+        $handler = Container::getInstance()->make(Handler::class);
+        try {
+            foreach ($this->exceptions as $exception) {
+                // hashed to be "anything other than 'production'
+                Config::set('app.env', md5('production'));
```
katjam: OK - but probs doesn't need to be so extreme. :)

charlesstrange2: You're right, could have been just `local`

PR #341, opened 2019-12-09
1.8/1440 upgrade libraries and deps

Nikomus:
https://trello.com/c/4oT2KJv3/1440-upgrade-libraries-and-dependencies-vv-e

- Moved store's deps into package.json
- Nothing seems to have broken, page weight should be very similar

image


resources/views/store/manage_vouchers.blade.php
```diff
@@ -2,10 +2,6 @@
 
 @section('title', 'Voucher Manager')
 
-@section('hoist-head')
```
charlesstrange2: Aha! good spot

PR #342, opened 2019-12-09
1.8/1447 new class of data worker

charlesstrange2:
Basic database and model changes to support further work. 

https://trello.com/c/azSNt5Oz/1447-new-class-of-data-worker-vv-ee


app/CentreUser.php
```diff
@@ -110,7 +122,7 @@ protected function homeCentres()
     /**
      * Get the relevant centres for this CentreUser, accounting for it's role
      *
-     * @return \Illuminate\Database\Eloquent\Collection|\Illuminate\Support\Collection|static[]
+     * @return Collection|\Illuminate\Support\Collection|static[]
```
Nikomus: if we're using `use` at the top of the file consistently, should we `use .. as ..` the base collection?

tests/Unit/Models/CentreUserModelTest.php
```diff
@@ -36,6 +38,22 @@ public function testCentreUserCanHaveNotes()
...
     }
 
+    /**@test */
+    public function testCentreUserCanHaveDownloadTrue()
+    {
+        // Standard CU
+        $cu = $this->centreUser;
+        $this->assertFalse($cu->downloader);
+
+        // Change his settings
```
Nikomus: Maybe "their" 🙂

PR #345, opened 2019-12-11
1.8/1450 update reg sheet download dob for cc data

charlesstrange2:
https://trello.com/c/23qxYpdd/1450-update-reg-sheet-download-dob-for-cc-data-users-v-e


charlesstrange2:
Closing to repopen and rebuild travis

charlesstrange2:
reopening to fiddle travis

app/Http/Controllers/Store/CentreController.php
```diff
@@ -50,20 +52,51 @@ public function printCentreCollectionForm(Centre $centre)
...
+        if (isset($centre->id)) {
+            // Set for specified centre
+            $centre_ids = [$centre->id];
+            $dateFormats = [
+                'dob' => 'm/Y'
+            ];
+        } else {
+            // Set for relevant centres
+            $centre_ids = $user->relevantCentres()->pluck('id')->all();
+            $dateFormats = [];
```
katjam: Is this because multiple centres means it's an ARC user download - not one that needs pretty dates?

katjam: Ah no - I see there are many date formats...

katjam: `$dateFormats` defaults to `[]` do you need to pass it?

app/Http/Controllers/Store/CentreController.php
```diff
@@ -50,20 +52,51 @@ public function printCentreCollectionForm(Centre $centre)
...
+    }
+
+    /**
+     * Returns array of formatted data about the Centre's Registrations
+     *
+     * @param array $centre_ids
+     * @param array $dateFormats
+     * @return array
+     */
+    private function getCentreRegistrationsSummary(array $centre_ids, $dateFormats = [])
```
katjam: Does anything else use this function? Maybe you don't need the `$dateFormats` param?

app/Http/Controllers/Store/CentreController.php
```diff
@@ -82,21 +115,19 @@ public function exportRegistrationsSummary()
...
         $headers = [];
 
         // Per registration...
         foreach ($registrations as $reg) {
             $lastCollection = $reg->bundles()
                 ->whereNotNull('disbursed_at')
                 ->orderBy('disbursed_at', 'desc')
-                ->first()
-            ;
+                ->first();
```
katjam: New auto-format rules?

app/Http/Controllers/Store/DashboardController.php
```diff
@@ -25,6 +25,7 @@ public function index()
             URL::route('store.registrations.print');
 
         $data = [
+            "user" => $user,
```
katjam: If you are passing whole `$user`, do you still need `$user->name`?

database/factories/ModelFactory.php
```diff
@@ -71,6 +68,23 @@
...
+$factory->defineAs(App\CentreUser::class, 'FMUser',function (Faker\Generator $faker) use ($factory){
+    static $password;
+
+    return [
+        'name' => $faker->name,
+        'email' => $faker->unique()->safeEmail,
+        'password' => $password ?: $password = bcrypt('secret'),
+        'remember_token' => str_random(10),
+        'role' => 'foodmatters_user',
+        'downloader' => true
```
katjam: Do you also have `foodmatters_users` who cannot download?

routes/store.php
```diff
@@ -141,36 +141,46 @@
...
+    ])
+        // Can view, can print individual form.
+        ->middleware('can:view,registration');
 
     // Batch print Family Forms for User Centre
     Route::get('/registrations/print', [
         'as' => 'store.registrations.print',
         'uses' => 'RegistrationController@printBatchIndividualFamilyForms',
     ]);
+    // If we keep this it needs a guard.
```
katjam: Which thing does the comment refer to?

app/Http/Controllers/Store/CentreController.php
charlesstrange2: I'd at least need to initialise it as null, as further down i just stuff the function with params.

charlesstrange2:
> Sad because many changes not documented in the trello ticket and only tangentially related.
> 
> But looks like you had a good tidy up... and nothing to stop approval. Couple of comments only.
> 
> Due to length of PR - I only skim read tests - but they look ok.

Yeah, tests in ARC are always behind features :-( i tried to cover the main points though

routes/store.php
charlesstrange2: The batch print famly route; Turns out another card in this sprint will remove it.

database/factories/ModelFactory.php
charlesstrange2: At this point _no_ but there's a possible thought that we might in the future.

app/Http/Controllers/Store/DashboardController.php
charlesstrange2: entirely for convenience and clarity in the blade; there's an include to make the top-right box with the current user name that requires it, and is included everywhere. I didn't want to have to check a whole bunch of blades at this point

app/Http/Controllers/Store/CentreController.php
charlesstrange2: No, I think my fault.

app/Http/Controllers/Store/CentreController.php
charlesstrange2: The decision about whihc formats is done in the controller function. I was considering moving this to registration, one reason i pulled it out into a function, the other being purpose separation.

PR #346, opened 2019-12-13
1.8/1429 cc worker can easily see non active families vvv ee

charlesstrange2:
https://trello.com/c/VPYJaqmH/1429-cc-worker-can-easily-see-non-active-families-vvv-ee

### DIFF
--- a/app/Http/Controllers/Store/CentreController.php
+++ b/app/Http/Controllers/Store/CentreController.php
@@ -191,15 +191,28 @@ class CentreController extends Controller
             $rows[] = $row;
         }
 
-        usort($rows, function ($a, $b) {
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Store/CentreController.php
```diff
@@ -187,19 +193,33 @@ private function getCentreRegistrationsSummary(array $centre_ids, $dateFormats =
...
             }
-            // stack new row onto the array
-            $rows[] = $row;
+
+            // Remove any keys we don't want and add to the list.
+            $rows[] = array_diff_key($row, $excludeColumns);
         }
 
-        usort($rows, function ($a, $b) {
+        // Sort te columns
```
katjam: the

PR #357, opened 2020-01-03
Update CSS to use fontawesome and shuffle some x/y values

charlesstrange2:
We were letting the browser decide what to use for a tick, this is now forced.

https://trello.com/c/QkyIGo2l/1466-edge-displaying-a-green-tick

### DIFF
--- a/public/store/css/main.css
+++ b/public/store/css/main.css
@@ -121,8 +121,8 @@ input[type="radio"]:checked+label:after,
 	position: absolute;
 	width: 0;
 	height: 0;
-	left: 9px;
-	top: 9px;
+	left: 10px;
+	top: 10px;
 	cursor: pointer;
 }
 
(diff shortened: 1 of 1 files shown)


public/store/css/main.css
```diff
@@ -132,9 +132,10 @@ input[type="radio"]:checked+label:after {
 }
 
 .styled-checkbox:checked+label:after {
-	content: "✔";
-	top: 8px;
-	left: 12px;
+	font-family: FontAwesome, 'Helvetica Neue', Helvetica, Arial, sans-serif;
+	content: "\f00c";
```
katjam: Does that code work in the other fonts?

katjam: Otherwise I think remove them as fallbacks or add as a more standard unicode char? https://www.fileformat.info/info/unicode/char/2714/index.htm

charlesstrange2: We did use that, it was the issue - the heavy check mark was showing as a weird green tick in edge, seemingly regardless of the font-family setting. 

Forcing it to use FontAwesome's icon is both more in keeping with the other icons we use and consistent across browsers. If FontAwesome fails to load lots of other icons will be broken.

charlesstrange2: I've removed the excess, pointless members. PhpStorm seems to think we need a default standard font-family too, hence sans-serif, to fall back on, but i don't expect it to ever be used.

PR #359, opened 2020-01-07
1.8/1475-pregnancy-date-shows-as-a-negative-figure-until-saved

charlesstrange2:
Add check to switch out P instead of yrs/mos if child is in the future when added.

https://trello.com/c/Rj2h9Ogh/1475-pregnancy-date-shows-as-a-negative-figure-until-saved

### DIFF
--- a/resources/views/store/partials/add_child_form.blade.php
+++ b/resources/views/store/partials/add_child_form.blade.php
@@ -88,9 +88,14 @@
                 var innerTextDate = dateObj.format("MMM YYYY");
                 var valueDate = dateObj.format("YYYY-MM");
 
-                // Make some age display values
-                var displayMonths = moment().diff(valueDate, 'months') % 12;
-                var displayYears = moment().diff(valueDate, 'years');
+                // Make some age display values based on if it's the future.
+                var ageString = (moment().diff(valueDate, 'days') > 0)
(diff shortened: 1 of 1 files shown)


resources/views/store/partials/add_child_form.blade.php
```diff
@@ -88,17 +88,22 @@ function () {
...
-                // Make some age display values
-                var displayMonths = moment().diff(valueDate, 'months') % 12;
-                var displayYears = moment().diff(valueDate, 'years');
+                // Make some age display values based on if it's the future.
+                var ageString = (moment().diff(valueDate, 'days') > 0)
+                    ? '<td class="age-col">' +
+                        moment().diff(valueDate, 'years') + ' yr, ' +
+                        moment().diff(valueDate, 'months') % 12  +
+                        ' mo</td>'
+                    : '<td class="age-col">P</td>'
```
katjam: What's the P?

Nikomus: It indicates pregnancy.

charlesstrange2: Having negative dates to represent unborn kids looked "wierd" to CC users, and hard to pick out. We went with "P" as they understood that better.

PR #360, opened 2020-01-10
1486-bug-fm-users-can-overwrite-verified-registrations-checked-value-…

charlesstrange2:
…unintentionally

Changed it so that verifying checkboxes are dependent on the registration context in edit, and the user's context in create.

https://trello.com/c/Tj5DwCPq/1486-bug-fm-users-can-overwrite-verified-registrations-checked-value-unintentionally

### DIFF
--- a/app/Http/Controllers/Store/RegistrationController.php
+++ b/app/Http/Controllers/Store/RegistrationController.php
@@ -183,7 +183,7 @@ class RegistrationController extends Controller
         // Grab carers copy for shift)ing without altering family->carers
         $carers = $registration->family->carers->all();
 
-        // Check if er verify, based on the existing registration's context
+        // Check if we verify, based on the existing registration's context
         $verifying = in_array($registration->centre->sponsor->shortcode, config('arc.verifies_children'));
 
         return view('store.edit_registration', array_merge(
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Store/RegistrationController.php
```diff
@@ -177,6 +183,9 @@ public function edit(Registration $registration)
         // Grab carers copy for shift)ing without altering family->carers
         $carers = $registration->family->carers->all();
 
+        // Check if er verify, based on the existing registration's context
```
katjam: er

resources/views/store/edit_registration.blade.php
```diff
@@ -138,7 +138,7 @@ class="@if($errors->has('pri_carer'))invalid @endif" type="text"
                                 {{--
                                 TODO : This is not scalable; if this breaks or has to be extended, look at a better abstraction.
                                 --}}
-                                @if (in_array($registration->centre->sponsor->shortcode, config('arc.extended_sponsors')) )
+                                @if ( $verifying )
                                     <li>Primary school to secondary school start - 3 vouchers<br>
                                         <small>(Only if family has a child still under primary school age too)</small>
```
katjam: Is this connected to verifying?

app/Http/Controllers/Store/RegistrationController.php
charlesstrange2: fixed

resources/views/store/edit_registration.blade.php
charlesstrange2: doh! reverted that one

PR #362, opened 2020-01-16
1.8/HOTFIX-1499-if-there-are-no-registrations-we-explode-printing-deshboard-reg-form

charlesstrange2:
We aren't catching a case where centres have no registrations for printing from the dashboard.

This is low priority, it's not likely in live.


charlesstrange2:
for some reason i think i git added my deploy builder and locks.
Probably useful to be there.

PR #367, opened 2020-02-17
1.9/Add-tests-1500-hotfix-permissions-issue-when-downloading-pdf-forms

AlinChindea:
Add tests for the functionality listed in this card - [https://trello.com/c/PKwBrsfX/1500-hotfix-permissions-issue-when-downloading-pdf-forms-e-vvv](https://trello.com/c/PKwBrsfX/1500-hotfix-permissions-issue-when-downloading-pdf-forms-e-vvv)

### DIFF
--- a/tests/Unit/Routes/StoreRoutesTest.php
+++ b/tests/Unit/Routes/StoreRoutesTest.php
@@ -41,6 +41,8 @@ class StoreRoutesTest extends StoreTestCase
     /** @var Centre $unrelatedCentre */
     private $unrelatedCentre;
 
+    private $dashboardRoute;
+
     public function setUp()
     {
         parent::setUp();
@@ -90,6 +92,8 @@ class StoreRoutesTest extends StoreTestCase
             "password" => bcrypt('test_unrelated_pass'),
         ]);
         $this->unrelatedUser->centres()->attach($this->unrelatedCentre->id, ['homeCentre' => true]);
+
+        $this->dashboardRoute = URL::route('store.dashboard');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tests/Unit/Routes/StoreRoutesTest.php
```diff
@@ -386,7 +388,7 @@ public function testCentreRegistrationsSummaryGate()
             "centre_id" => $this->centre->id,
         ]);
 
-        $route = $specificCentreRoute = URL::route('store.centre.registrations.summary', ['centre' => $this->centre->id]);
```
charlesstrange2: cant recall why i had it set that way :thinking:

AlinChindea: don't know 🤔 but removing it doesn't break anything. there's also an unused declaration `use Illuminate\Support\Collection;` that we can remove from this file, I think.

charlesstrange2: i'm using that to specifically highlight $registrations as collection; on like 21.
It helps with autocomplete in IDE's
Can be removed.

tests/Unit/Routes/StoreRoutesTest.php
```diff
@@ -527,11 +528,20 @@ public function testMVLRouteGuard()
             ->get($route)
             ->assertResponseStatus(403)
         ;
+
+        Auth::logout();
+
+        // Throw a 403 for auth'd but forbidden
+        $this->actingAs($this->downloaderUser, 'store')
```
charlesstrange2: aha! yes, good spot

AlinChindea:
> Optionally remove references to the Collection class, as you suggested
> otherwise 👍

Removing that unused declarations somehow breaks a test - [see here](https://travis-ci.org/neontribe/ARCVService/builds/651594183?utm_source=github_status&utm_medium=notification) - so I added it back. Merging.

PR #368, opened 2020-02-17
1.9/1499-hotfix-if-there-are-no-registrations-we-explode-when-printing-dashboard-collections-form

AlinChindea:
Add tests for this card [1.9/1499-hotfix-if-there-are-no-registrations-we-explode-when-printing-dashboard-collections-form](1.9/1499-hotfix-if-there-are-no-registrations-we-explode-when-printing-dashboard-collections-form)


yarn.lock
```diff
@@ -938,11 +938,6 @@
   resolved "https://registry.yarnpkg.com/@xtuc/long/-/long-4.2.2.tgz#d291c6a4e97989b5c61d9acf396ae4fe133a718d"
```
charlesstrange2: Do we need yarn.lock - is it because of this morning's breaking drive?

AlinChindea: as discussed, we keep it in for now

PR #372, opened 2020-02-20
1.9/1509 southwark seeing is id seen box ticked in a rapid way during voucher processing ee vv

charlesstrange2:
Create a new family level notice that applies to SK-like places that use the extended rules. 
At some point we may need composable dynamic rules sets, but not today.

The reason that a lot of the notices weren't showing at all was most likely the Valuation::flat() function was broken and wasn't collating them. This is now fixed.

Fixed the notifications box in column 1 - it now fills properly

Added a test to check that we can see family based notifications, at this point that it has unverified children or is about to have children drop off the service.

Modernised the registration factory to take parameters if it needs to.


app/Services/VoucherEvaluator/Evaluations/FamilyHasUnverifiedChildren.php
```diff
@@ -0,0 +1,49 @@
...
+
+        $satisfiers = array_filter(
+            $children,
+            function ($child) {
+                // If we satisfy the rule
+                return ($this->specification->isSatisfiedBy($child));
+            }
+        );
+
+        if (empty($satisfiers)) {
```
AlinChindea: I keep seeing people checking for empty arrays with the `$empty()` or the `count($array) < 1` or `isset()`. Are there any advantages of one way over another? Or it's just down to personal preference? I'll look deeper into it - I did already but my memory is not that good - but am also curious about your choice.

app/Services/VoucherEvaluator/Valuation.php
```diff
@@ -133,28 +133,39 @@ public function getEligibility()
...
      * @return mixed|null
      */
-    public function flat($attribute, bool $onlyEligible = false)
+    public function flat($attribute, bool $onlyEligibility = null)
     {
-        // Check this attribute
-        if (property_exists($this, $attribute) &&
-            // If this valuation isn't eligible
-            $this->getEligibility() === $onlyEligible
+
```
AlinChindea: Whitespace. don't know if you want to leave an empty line betwee the opening curly brace and the first line of code/comment.

database/factories/ModelFactory.php
```diff
@@ -327,23 +327,41 @@
...
+    ;
 
     // Set initial centre (and thus, rvid)
     $family->lockToCentre($centre);
     $family->save();
 
     // Add dependent models
-    $family->carers()->saveMany(factory(App\Carer::class, random_int(1, 3))->make());
-    $family->children()->saveMany(factory(App\Child::class, random_int(0, 4))->make());
+    if ($family->carers->count() == 0) {
```
AlinChindea: Don't you want strict equality here?

database/factories/ModelFactory.php
```diff
@@ -327,23 +327,41 @@
...
+        $family->carers()->saveMany(factory(App\Carer::class, random_int(1, 3))->make());
+    }
+
+    if ($family->children->count() == 0) {
```
AlinChindea: again, strict equality maybe? ☝️

resources/views/store/manage_vouchers.blade.php
```diff
@@ -32,11 +32,10 @@
                     <div class="icon-container warning">
                         <i class="fa fa-exclamation-triangle" aria-hidden="true"></i>
                     </div>
-                    <div>
+                    <div id="family-warning">
                         @foreach( $noticeReasons as $notices )
                             <p class="v-spaced">
-                                Warning: {{ $notices['count'] }} {{ str_plural($notices['entity'], $notices['count']) }}
-                                is {{ $notices['reason'] }}
+                                Warning: {{ $notices['count'] }} {{ str_plural($notices['entity'], $notices['count']) }} currently "{{ $notices['reason'] }}"
```
AlinChindea: Is this the bit where it should be `Reminder`? if so, maybe update the id as well.

tests/Feature/Store/VoucherManagerTest.php
```diff
@@ -90,6 +89,75 @@ public function testThreeColumnsAreVisible()
...
+            'homeCentre' => true
+        ]);
+
+        // Navigate to the page as that user.
+        $this->actingAs($ccUser, 'store')
+            ->visit(URL::route('store.registration.voucher-manager', [ 'id' => $registration ]))
+        ;
+
+        // We can see the warnings area
+        $this->seeElement("#family-warning");
```
AlinChindea: this might be `reminder` if changed in code

AlinChindea:
Left some comments/questions for you. Rather to double check some things. Also, maybe it's better to add the link to the Trello card in the PR description above? 🤷‍♂

app/Services/VoucherEvaluator/Evaluations/FamilyHasUnverifiedChildren.php
charlesstrange2: isset() will ask if the variable is null, _or even declared_. It exists because you sometimes deep objects and array structures are lazily populated.

empty is a language construct, not really a function. It's a bit more flexible than isset(), in that IIRC, it can spot that an array *exists*, but has no content - see https://www.php.net/manual/en/function.empty for the difference.

count($array) assumes there _is_ an array to count.

I use empty becuase i'm a little lazy, and it seems to catch conditions that i would define as "not present, or not useful"

app/Services/VoucherEvaluator/Valuation.php
charlesstrange2: True, will remove

database/factories/ModelFactory.php
charlesstrange2: yep!

database/factories/ModelFactory.php
charlesstrange2: yep!

resources/views/store/manage_vouchers.blade.php
charlesstrange2: This will change in the card Kat and Faith are looking at.

tests/Feature/Store/VoucherManagerTest.php
charlesstrange2: Could probably be "family-notice", will change in next card

database/factories/ModelFactory.php
```diff
@@ -327,23 +327,41 @@
...
 
     $eligibilities = ['healthy-start', 'other'];
 
-    $centre = App\Centre::inRandomOrder()->first();
+    if (!empty($attributes['centre_id'])) {
+        // Use the passed centre id.
+        $centre = App\Centre::find($attributes['centre_id']);
+    } else {
+        // Default to a random centre.
+        $centre = App\Centre::inRandomOrder()->first();
```
katjam: neat trick

tests/Feature/Store/VoucherManagerTest.php
```diff
@@ -73,11 +76,7 @@ public function setUp()
         Auth::logout();
     }
 
-    /**
```
katjam: lol

tests/Feature/Store/VoucherManagerTest.php
```diff
@@ -90,6 +89,75 @@ public function testThreeColumnsAreVisible()
...
+        foreach ($notices as $notice) {
+            if ($notice["entity"] == "Family") {
+                $this->see($notice["reason"]);
+            } else {
+                $this->dontSee($notice["reason"]);
+            }
+        }
+    }
+
+    /** @test */
```
katjam: do you need this if the test function name starts with `test`? Why mixing naming conventions?

charlesstrange2: There have been occasions when phpunit has skipped withiut the directive :man_shrugging: at the moment its witchcraft but im filling in places i missed it out.

PR #373, opened 2020-02-21
1.9/1517 Date of file Downloaded column added to registration file for FM Users

AlinChindea:
## Trello card - [https://trello.com/c/lmB5xtkT/1517-date-of-data-column-for-reg-data-download-e-vv](https://trello.com/c/lmB5xtkT/1517-date-of-data-column-for-reg-data-download-e-vv)


app/Http/Controllers/Store/CentreController.php
```diff
@@ -106,6 +106,10 @@ public function exportRegistrationsSummary(Centre $centre)
      */
     private function getCentreRegistrationsSummary(array $centre_ids, $dateFormats = [], $excludeColumns = [])
     {
+        // Get User
+        /** @var CentreUser $user */
+        $user = Auth::user();
```
charlesstrange2: This works, but may be unnecessary, as the user is checked for the permissions above in `exportRegistrationsSummary`

We can pass "Date file was Downloaded" as a value in the "excludeColumns" array.
Then we can check if we need to add the date to the array further down

app/Http/Controllers/Store/CentreController.php
```diff
@@ -201,6 +205,10 @@ private function getCentreRegistrationsSummary(array $centre_ids, $dateFormats =
             // Would be confusing if an old reason was left in - so check leaving date is there.
             $row["Leaving Reason"] = $reg->family->leaving_on ? $reg->family->leaving_reason : null;
 
+            if ($user->role === "foodmatters_user") {
```
charlesstrange2: This check may not be necessary, if we pass "Date file was downloaded" in $excludeColumns

app/Http/Controllers/Store/CentreController.php
AlinChindea: I've amended my code. However, I was a bit a confused about your remark:
> Then we can check if we need to add the date to the array further down

app/Http/Controllers/Store/CentreController.php
```diff
@@ -193,21 +190,19 @@ private function getCentreRegistrationsSummary(array $centre_ids, $dateFormats =
...
+            $row['Join Date'] = $reg->created_at ? $reg->created_at->format($dateFormats['join']) : null;
+            $row['Leaving Date'] = $reg->family->leaving_on ? $reg->family->leaving_on->format($dateFormats['leave']) : null;
             // Would be confusing if an old reason was left in - so check leaving date is there.
-            $row["Leaving Reason"] = $reg->family->leaving_on ? $reg->family->leaving_reason : null;
+            $row['Leaving Reason'] = $reg->family->leaving_on ? $reg->family->leaving_reason : null;
 
-            if ($user->role === "foodmatters_user") {
-                $row["Date file was Downloaded"] = Carbon::today()->toDateString();
-            }
+            $row['Date file was Downloaded'] = Carbon::today()->toDateString();
```
charlesstrange2: This works absolutely fine - you could also, just skip if 'Date file was Downloaded' is a member of the excludeColumns array.

AlinChindea: we can skip it? but should it not be created first? asign it a date value? adding it to the `$excludeColumns` just creates that column if it doesn't exist? and if so, how would it know about its date value? unless we have to do something else in the `exportRegistrationsSummary()` and nothing in the `getRegistrationsSummary()` 🤔 😕

charlesstrange2: if the function has been told "skip showing this column" there's not much point in making it, just to get rid of it further down. 

unless i'm missing something?

AlinChindea: we do want to show that column to fm users, don't we? that's my understanding of this card. it excludes it for those who do not have the right to export the registrations. 
Screenshot 2020-02-24 at 17 32 50

charlesstrange2: Yes, that is correct.

PR #375, opened 2020-02-24
1.9/1507 Create UI for ARC to add data-downloader flage to selected CC workers

AlinChindea:


tests/Unit/Controllers/Service/Admin/CentreUserControllerTest.php
```diff
@@ -35,7 +35,8 @@ public function setUp()
             'name' => 'bobby testee',
             'email' => 'bobby@test.co.uk',
             'worker_centre' => $this->centre->id,
-            'alternative_centres.*' => $this->altCentres->pluck('id')->all()
+            'alternative_centres.*' => $this->altCentres->pluck('id')->all(),
+            'downloader' => 1,
```
charlesstrange2: set here, `$this->data` will be used in both new and update tests.
In the `testICanStoreACentreUser()` we should see(), or seeInDatabase() that the `'downloader' => 1` has _not_ affected_ the default CentreUser construction of `0`. That way, we test that CentreUsers are created without that power.

tests/Feature/Service/ServiceWorkersPageTest.php
```diff
@@ -0,0 +1,45 @@
...
+        $this->adminUser = factory(AdminUser::class)->create();
+        $this->workersRoute = route('admin.centreusers.index');
+    }
+
+    /**
+     * @test
+     *
+     * @return void
+     */
+    public function itShowsAListOfWorkers()
```
charlesstrange2: This shows a table with headers, but not necessarily users in that list.

I'd suggest setUp() creates a 
- 3 centres in an area
- 1 user with with a homeCentre and no alternatives
- 1 user with with a homeCentre and 2 alternatives
- 1 user with a homeCentre who can Download.

StoreTestCase::seeInElementAtPos() is a custom function available like seeElement() but it can check that elements are visible in a set of elements that are pulled out of a page using a selector. We can use it to check that an array of elements has specific element and contents at an index.

AlinChindea: @charlesstrange2 I've updated this test based on your feedback :point_up: Can you have a look and let me know if it's ok? You might need to re-approve the PR. :croissant:

tests/Feature/Service/ServiceWorkersPageTest.php
```diff
@@ -5,29 +5,73 @@
...
+
     private $workersRoute;
 
     public function setUp()
     {
         parent::setUp();
+
         $this->adminUser = factory(AdminUser::class)->create();
+        // Create 3 centres in an area
+        $this->centre = factory(Centre::class, 3)->create();
```
charlesstrange2: Need a sponsor, then bind each centre to the sponsor,

you can use `->each(function($c) {... }); `

tests/Feature/Service/ServiceWorkersPageTest.php
```diff
@@ -5,29 +5,73 @@
...
+            "email" => "testuser@example.com",
+        ]);
+        $this->userHomeCentreNoAlt->centres()->attach(1, ['homeCentre' => true]);
+
+        // Create 1 user with with a homeCentre and 2 alternatives
+        $this->userHomeCentreTwoAlt = factory(CentreUser::class)->create([
+            "name"  => "test user with alternatives",
+            "email" => "testuseralternatives@example.com",
+        ]);
+        $this->userHomeCentreTwoAlt->centres()->attach($this->altCentres->pluck('id')->all(), ['homeCentre' => true]);
```
charlesstrange2: you've set homeCentre True on all centres. technically possible but dangerous, a user should only have one homeCentre.

1. create the user
2. add a single centre, with homeCentre=true
3. add a couple of atlernatives, without homeCentre

PR #391, opened 2020-03-13
1.9/1545-safari-styling-required-on-search-for-a-family-page

AlinChindea:

### DIFF
--- a/public/store/css/main.css
+++ b/public/store/css/main.css
@@ -907,10 +907,13 @@ SEARCH PAGE
 	outline-offset: 0px;
 }
 
+.checkbox-control.relative {
+	display: flex;
+}
+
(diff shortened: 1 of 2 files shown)


public/store/css/main.css
```diff
@@ -907,10 +907,13 @@ SEARCH PAGE
 	outline-offset: 0px;
 }
 
+.checkbox-control.relative {
```
Nikomus: .relative is a helper class that only applies `position: relative;` so I would just remove the class in the blade, and add the `display: flex` in the block below.

charlesstrange2: I bow to nick's css wisdom

AlinChindea: very well! shall do.

PR #398, opened 2020-05-13
1.10/1403-write-tests

AlinChindea:
### Trello card 👉 [https://trello.com/c/5bI5n2Ei/1403-write-tests-ee-vv](https://trello.com/c/5bI5n2Ei/1403-write-tests-ee-vv)


tests/Unit/Routes/StoreRoutesTest.php
```diff
@@ -457,10 +430,61 @@ public function testSessionUpdateGate()
...
+            $this->centreUser->centre->sponsor->id
         );
+
+        try {
+            $this->actingAs($this->unrelatedUser, 'store')
+                ->visit(URL::route('store.registration.edit', $registration->id))
+                ->call(
+                    'PUT',
+                    $route,
+                    []
```
katjam, line 483: `['leaving_reason' => 'found employment']
￼         `

tests/Unit/Routes/StoreRoutesTest.php
```diff
@@ -457,10 +430,61 @@ public function testSessionUpdateGate()
...
+            ->seePageIs(URL::route('store.login'))
+            ->assertResponseStatus(200);
+
+        $this->actingAs($this->centreUser, 'store')
+                ->visit($route)
+                ->assertResponseStatus(200);
+
+        $this->actingAs($this->centreUser, 'store')
+            ->visit(URL::route('store.registration.edit', $registration->id))
+            ->press('Remove this family')
```
katjam, line 451: Don;t need the button press

app/Http/Controllers/Service/UserController.php
```diff
@@ -18,69 +17,15 @@ public function index()
...
-    }
-
     /**
      * Display the specified resource.
      *
+     * @codeCoverageIgnore
      * @param  \App\User  $user
      * @return \Illuminate\Http\Response
      */
     public function show(User $user)
```
katjam, line 27: Why are we leaving this in?

app/Http/Controllers/API/TraderController.php
```diff
@@ -183,7 +147,8 @@ public function emailVoucherHistory(Request $request, Trader $trader)
         // If a date is provided generate a specific response message.
         if ($date) {
             $response_text = trans(
-                'api.messages.email_voucher_history_date', [
+                'api.messages.email_voucher_history_date',
+                [
```
katjam, line 151: Not sure about some of these syntax changes - but I know it's your editor. No harm done, but we might want to think about normalising a little.

app/Http/Controllers/Store/BundleController.php
```diff
@@ -52,8 +51,7 @@ public function create(Registration $registration)
         $lastCollection = ($lastCollectedBundle && (!empty($lastCollectedBundle->disbursed_at)))
             // 'disbursed_at' is auto-carbon'd by the Bundle model
             ? $lastCollectedBundle->disbursed_at->format('l jS \of F Y')
-            : null
-        ;
+            : null;
```
katjam, line 54: @charlesstrange2 Which do you prefer? The ARC codebase seems to have both.

app/Http/Controllers/Service/UserController.php
charlesstrange2: It can probably go

app/Http/Controllers/Store/BundleController.php
charlesstrange2: I genuinly have no preference generally. I tend to do it when i have multiline call chains
```
$thing->select()
  ->where('stuff')
  ->get()
  ->sortBy('sorting_thing')
;
```

app/Providers/EventServiceProvider.php
```diff
@@ -13,9 +13,6 @@ class EventServiceProvider extends ServiceProvider
      * @var array
      */
     protected $listen = [
-        'App\Events\Event' => [
-            'App\Listeners\EventListener',
```
charlesstrange2, line 17: yep, this is pointlesss

tests/Unit/Routes/DataRoutesTest.php
```diff
@@ -37,19 +36,17 @@ public function testVouchersIndexRouteAuthdAdmin()
         $this->actingAs($this->admin, 'admin')
             ->get(route('vouchers.index'))
             ->assertStatus(200)
-            ->assertJsonStructure([ 0 => [
+            ->assertJsonStructure([0 => [
                 'id', 'trader_id', 'code', 'currentstate', 'sponsor_id'
-            ]])
-        ;
+            ]]);
```
charlesstrange2, line 41: This is an example of where i would favour the next-line `;` 
Not important though

katjam: I also favour the next line `;` syntax, but agree it is not important. The reason I bring it up is:

1- Much time has been spent by me and others discussing. @Nikomus points out that there are examples of both styles in this project so it's confusing to know which to go for
2- Some time spend in Code Review looking at the irrelevantly changed lines

katjam: I don't think it's worth the time to change these back now, but it might be worth agreeing that we do closing `;` on their own line from now on and configuring our editors not to undo this preference in future.

AlinChindea: I would very much prefer a clear coding standard, especially if there is nothing imposed by the client. I know it would be per project but at least then we're all on the same page

Nikomus: Lonely `;` looks weird to me, but more importantly I think it can be demoralising to hold up a PR fiddling with these when it's not a framework convention, or well observed in the project code/defined in docs.

app/Http/Controllers/Store/BundleController.php
AlinChindea: @charlesstrange2 do you want me to reverse these in the multiline call, i.e. tests especially in this PR?

PR #403, opened 2020-05-21
1.10/1564 remove printed and ordered

katjam:
https://trello.com/c/qjEN0UY6/1564-remove-printed-and-ordered-states-from-voucher-transitions-there-was-a-bug-so-150k-didnt-have-them-recorded-anyway-and-they-are

- Remove requested and ordered states from system
- Initial voucher state is now printed

### DIFF
--- a/database/migrations/2020_05_21_094708_remove_ordered_printed_from_voucher_states_table.php
+++ b/database/migrations/2020_05_21_094708_remove_ordered_printed_from_voucher_states_table.php
@@ -1,0 +1,39 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class RemoveOrderedPrintedFromVoucherStatesTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('voucher_states', function (Blueprint $table) {
+            // I guess nice to have a record of how many rows were deleted?
+            $number_to_delete = App\VoucherState::whereIn('to', ['ordered', 'printed'])->count();
+
+            App\VoucherState::whereIn('to', ['ordered', 'printed'])->delete();
+            Log::info($number_to_delete . ' printed and ordered voucher_states records deleted on migration.');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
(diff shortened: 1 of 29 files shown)


database/migrations/2020_05_21_094708_remove_ordered_printed_from_voucher_states_table.php
```diff
@@ -0,0 +1,39 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('voucher_states', function (Blueprint $table) {
+            // I guess nice to have a record of how many rows were deleted?
+            $number_to_delete = App\VoucherState::whereIn('to', ['ordered', 'printed'])->count();
+
+            App\VoucherState::whereIn('to', ['ordered', 'printed'])->delete();
+            Log::info($number_to_delete . ' printed and ordered voucher_states records deleted on migration.');
```
AlinChindea, line 21: You want to leave this `Log::info()` in?

katjam: The log is for historical info/ record. We do it in loads of places so we can easily look into activity in production. In particular around destructive or suspicious actions.

If you search the code base for `Log::info(` you'll see the types of things we log.

katjam: It doesn't. I think if you aren't in App Laravel prefers the Log import not to be there.

You get that error when importing to tests.

database/migrations/2020_05_21_094708_remove_ordered_printed_from_voucher_states_table.php
```diff
@@ -0,0 +1,39 @@
...
+class RemoveOrderedPrintedFromVoucherStatesTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('voucher_states', function (Blueprint $table) {
```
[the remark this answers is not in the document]
katjam: I think we need top run it everywhere in case people do a migrate rather than a migrate:refresh --seed in any other environs.

database/migrations/2020_05_21_094708_remove_ordered_printed_from_voucher_states_table.php
katjam: The 'has no effect' error. I have run this migration and rolled back multiple times on my local to check stuff.

PR #404, opened 2020-05-21
1.10/1487 improve live staging parity seeds 

AlinChindea:
### Seeds here 👉 [https://trello.com/c/dspZteO3/1579-set-up-test-cases-for-may-2020-sprint](https://trello.com/c/dspZteO3/1579-set-up-test-cases-for-may-2020-sprint)

❗**NB**: Collection history not included (done in #406)


database/seeds/RegistrationsSeeder.php
```diff
@@ -37,9 +37,120 @@ public function run()
         // create a registration with a bundle
         $bundle = factory(App\Bundle::class)->create();
         factory(App\Registration::class)->create()->bundles()->save($bundle);
-    }
+
+        // centre without SK rules for fam1 & fam4 that is not in same area as families 3, 5 and 6
```
katjam, line 41: This won't be SK rules after this sprint, would it make sense to call it family rule? Or disqualifier rule? @charlesstrange2 thoughts?

database/seeds/RegistrationsSeeder.php
```diff
@@ -37,9 +37,120 @@ public function run()
...
+        $testCentreFam1And4 = App\Centre::find(7);
+        $testCentreFam1And4->sponsor_id = App\Sponsor::find(5)->id;
```
katjam, line 43: Love the names. Very clear.

katjam, line 43: I think you should get the sponsor id from the centre so that they are consistent with the data

        `$testCentreFam1And4->sponsor->id;`

database/seeds/RegistrationsSeeder.php
```diff
@@ -37,9 +37,120 @@ public function run()
...
+        $testCentreFam5->sponsor_id = App\Sponsor::find(3)->id;
+
+        /*
+        centre with non SK rules for fam 6 that is in a different area from all other families
+        TODO: check that we don't have SK rules
+        */
+        $testCentreFam6 = App\Centre::find(6);
+        $testCentreFam6->sponsor_id = App\Sponsor::find(4)->id;
+
+        $familyData1 = [
```
katjam, line 73: It might be clearer to have the `$testCentreFam1` passed in this array also, like

`
```
'center' => $testCentreFam1`,
'carers' => [[...]],
...
```

database/seeds/RegistrationsSeeder.php
AlinChindea: the reason I'm doing a re-assignment (a centre is attached to a sponsor by default it seems) is to make sure that each of these centres are in a different area (that is one of the requests on the ticket). if I just the centre's sponsor id we run the risk of having 2 or more centres in the same areas because this is randomly assigned. at least that's my understanding so correct me if I'm wrong. 🧇

katjam: It might be better to set up the centres in their seeds instead?

katjam: I don't think you should be altering them here.

katjam: Especially since you don't save the centre after. It will make the data inconsistent.

AlinChindea: so then, not to use the existing centres but create new ones? again, is it right that creating a new centre assigns it to a sponsor/area _randomly_? if so, there's a chance that some of these test families will be in the same area and that's what we're trying to avoid here.

database/seeds/RegistrationsSeeder.php
charlesstrange2: Where you're checking child requires verification, use "child verification rules"
Where you're checking child qualification use "child qualification rules"

database/seeds/RegistrationsSeeder.php
charlesstrange2: Since a centre requires a sponsor to save itself, the factory 
- creates one if there are none
- randomly picks one of the existing ones if there's one or more

So, when using a factory you need to reassign the sponsor if it's not the one you want.
Then save the centre so the seed is persisted in DB

If you need a centre for a specific test, and one centre does not seem to be precisely appropriate, probably best to make a new one, as we can't garuantee that it wont cause strange issues if other seeds use the same centre.

database/seeds/RegistrationsSeeder.php
AlinChindea: how can I tell which one applies to which test family?

database/seeds/RegistrationsSeeder.php
AlinChindea: well, from all this I take that:
1) if I don't want any funny behaviour of the seeds these test families should have new centres and not use existing ones. So better to be defensive here. Question is: should I or not? I understand the argument but what would **you** do? I'd want to create new centres for these test families even though existing ones seem appropriate but there's an unquantifiable risk of side effects from other seeds using those centres. 
2) might be better to set the centres in the seeds - readability? or is there any other reason?

 If you want me to make a decision here, I will but it seems I made the wrong or sub-optimal decisions so far. Otherwise we wouldn't be having this discussion 😄

katjam: I think you are being a bit hard on yourself. My opinion. If the existing seeded centres are not consistent (i.e. random on generation) or not suitably configured, make new ones.

If the existing ones work for you, use them as they are or alter them when they get generated in their own seeder, rather than here.

AlinChindea: I moved the centres inside the seeds where these are created anew (with their own sponsor id) and the seeds are now in a private function. It's a cleaner look from my point of view,separating concerns. hope it's better now.

database/seeds/RegistrationsSeeder.php
```diff
@@ -77,4 +90,186 @@ public function createRegistrationForCentre($quantity, App\Centre $centre = null
         // Return the collection in case anyone needs it.
         return collect($registrations);
     }
+
+    /**
+     * @param array $familyData
+     * @param Centre $centre
+     * @return void
+     */
+    public function createRegistrationTestCase($familyData, App\Centre $centre = null)
```
charlesstrange2, line 99: optional : probably don't need the "App\" due to imports

PR #407, opened 2020-05-22
1.10/1551 void voucher error message with success

katjam:
https://trello.com/c/6Ep7bSZL/1551-vouchers-that-have-already-been-paid-or-recorded-for-payment-cannot-be-marked-as-void-specify-voucher-in-error-message-eee-v

- Allow batches of void vouchers to contain non-voidable.
- success details contain list of unvoidable vouchers.

There is certainly room for improvement, this is a first hack.

### DIFF
--- a/app/Http/Controllers/Service/Admin/DeliveriesController.php
+++ b/app/Http/Controllers/Service/Admin/DeliveriesController.php
@@ -93,7 +93,6 @@ class DeliveriesController extends Controller
 
                 // Bulk update VoucherStates for speed.
                 foreach ($transitions as $transitionDef) {
-                    DB::enableQueryLog();
                     Voucher::select('id')
                         ->whereNull('delivery_id')
                         ->inDefinedRange($rangeDef)
--- a/app/Http/Controllers/Service/Admin/DeliveriesController.php
+++ b/app/Http/Controllers/Service/Admin/DeliveriesController.php
@@ -96,7 +96,8 @@ class DeliveriesController extends Controller
                     DB::enableQueryLog();
                     Voucher::select('id')
                         ->whereNull('delivery_id')
-                        ->withRangedVouchersInState($rangeDef, 'printed')
+                        ->inDefinedRange($rangeDef)
+                        ->inOneOfStates(['printed'])
                         ->chunk(
                             // should be big enough chunks to avoid memory problems
                             10000,
--- a/app/Http/Controllers/Service/Admin/VouchersController.php
+++ b/app/Http/Controllers/Service/Admin/VouchersController.php
@@ -53,25 +53,31 @@ class VouchersController extends Controller
     }
 
     /**
-     * Update Voucher range - specifically, state
+     * Retire voucher range - through void or expire
      *
      * @param AdminUpdateVoucherRequest $request
      * @return RedirectResponse
      */
-    public function updateBatch(AdminUpdateVoucherRequest $request)
+    public function retireBatch(AdminUpdateVoucherRequest $request)
     {
-        // Make a rangeDef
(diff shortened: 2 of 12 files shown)


app/Http/Controllers/Service/Admin/VouchersController.php
```diff
@@ -53,26 +53,32 @@ public function void()
...
+        if ($voidableVouchers->count() === 0) {
+            // Whoops! None of the vouchers are voidable
             return redirect()
-                ->route('admin.vouchers.void')
-                ->withInput()
-                ->with('error_message', trans('service.messages.vouchers_batchtransition.blocked'));
+              ->route('admin.vouchers.void')
+              ->withInput()
+              ->with('error_message', trans('service.messages.vouchers_batchretiretransition.blocked'));
+            return;
```
AlinChindea, line 78: if there are, shouldn't we get a success message and display it to the user?

app/Providers/CustomValidationProvider.php
```diff
@@ -59,6 +59,34 @@ public function boot()
...
+            $val = Voucher::splitShortcodeNumeric($value);
+
+            // Grab the content of the parameter we pass the rule in a roundabout fashion
+            $otherVal = array_get(
+                $validator->getData(),
+                $parameters[0]
+            );
+
+            if (!empty($otherVal)) {
+                $other = Voucher::splitShortcodeNumeric($otherVal);
```
AlinChindea, line 76: we should be more descriptive with the variable names, no?

katjam: Was straight copy of the block above. If you decide to change, please change there also.

app/Voucher.php
```diff
@@ -538,22 +442,51 @@ public static function getMinMaxVoucherDates(Collection $vouchers)
...
+    /**
+     * Gets a set of vouchers that are voidable.
+     *
+     * @param Builder $query
+     * @param array $states
+     * @return Builder
+     */
+    public function scopeInVoidableState(Builder $query)
+    {
+        $voidable_states = ['dispatched'];
```
AlinChindea, line 486: shouldn't this include `printed`, too? Maybe am not fully getting it but above - line 180 you wrote that `$voidable_states = ['printed', 'dispatched'];`

app/Voucher.php
```diff
@@ -174,28 +175,21 @@ public static function createRangeDefFromVoucherCodes($startCode, $endCode)
...
     }
 
-    /**
-     * Determines if the given voucher range contains entries already delivered
-     *
-     * @param $rangeDef object { 'start', 'end', 'shortcode', 'sponsor_id' }
-     * @return bool
-     */
-    public static function rangeIsDeliverable($rangeDef)
+    public static function collectionHasVoidable($currentstates)
```
AlinChindea, line 178: this method doesn't seem to be used, neither are its argument `currentstates` and the declared variable (`$voidable_states`) 🤔

app/Voucher.php
charlesstrange2: We should only be voiding dispatched vouchers at this point

app/Http/Controllers/Service/Admin/DeliveriesController.php
```diff
@@ -96,7 +96,8 @@ public function store(AdminNewDeliveryRequest $request)
                     DB::enableQueryLog();
```
charlesstrange2, line 96: Whoops! this can go

app/Voucher.php
charlesstrange2: I'll prune it

app/Providers/CustomValidationProvider.php
charlesstrange2: done

PR #400, opened 2020-05-14
1.10/1563-ability-to-download-worker-list-with-a-click-of-a-button

AlinChindea:
### Trello card 👉 [https://trello.com/c/hs9uwODu/1563-ability-to-download-worker-list-with-a-click-of-a-button-ee-vv](https://trello.com/c/hs9uwODu/1563-ability-to-download-worker-list-with-a-click-of-a-button-ee-vv)

PR #411, opened 2020-06-01
1.10/Ability to download workers sorted by home centre name and then worker name

AlinChindea:
### Trello card 👉 [https://trello.com/c/hs9uwODu/1563-ability-to-download-worker-list-with-a-click-of-a-button-ee-vv](https://trello.com/c/hs9uwODu/1563-ability-to-download-worker-list-with-a-click-of-a-button-ee-vv)

❗ Only deals with one item on that card. Another PR dealt with the bulk of the logic of this card. See #400 

`sortBy()` multiple columns is an issue in Laravel - [https://github.com/laravel/ideas/issues/11](https://github.com/laravel/ideas/issues/11) so I took to googling and followed this suggestion from a [StackOverflow answer](https://stackoverflow.com/questions/25451019/what-is-the-syntax-for-sorting-an-eloquent-collection-by-multiple-columns/25451441#25451441)

### DIFF
--- a/app/Http/Controllers/Service/Admin/CentreUsersController.php
+++ b/app/Http/Controllers/Service/Admin/CentreUsersController.php
@@ -223,7 +223,10 @@ class CentreUsersController extends Controller
 
     public function download()
     {
-        $workers = CentreUser::get()->sortBy('homeCentre.name');
+        $workers = CentreUser::get()->sortBy(function($centreUser) {
+            return sprintf('%-12s%s', $centreUser->homeCentre->name, $centreUser->name);
+        });
+
         $csvExporter = new Export();
 
         /**
--- a/app/Http/Controllers/Service/Admin/CentreUsersController.php
+++ b/app/Http/Controllers/Service/Admin/CentreUsersController.php
@@ -223,9 +223,7 @@ class CentreUsersController extends Controller
 
     public function download()
     {
-        $workers = CentreUser::get()->sortBy(function($centreUser) {
-            return sprintf('%-12s%s', $centreUser->homeCentre->name, $centreUser->name);
-        });
+        $workers = CentreUser::get()->sortBy('name')->sortBy('homeCentre.name');
 
         $csvExporter = new Export();
 


app/Http/Controllers/Service/Admin/CentreUsersController.php
```diff
@@ -223,7 +223,10 @@ private function syncCentres($request, CentreUser $cu)
 
     public function download()
     {
-        $workers = CentreUser::get()->sortBy('homeCentre.name');
+        $workers = CentreUser::get()->sortBy(function($centreUser) {
+            return sprintf('%-12s%s', $centreUser->homeCentre->name, $centreUser->name);
```
charlesstrange2, line 227: ah! hashing the  two together and sorting them
I read that as 
parameter 1, left jsutified, first 12 characters,
Parameter 2, not justified. 

why the twelve characters, and why the left justify?

charlesstrange2: Also,
could you return 
"string1#string2"
using the delimeter "#"?

AlinChindea: I don't really get it what's going on here. But here goes an explanation I gleaned from teh interwebs. If you need the sortBy against multiple columns, you probably need to space pad them to ensure that "ABC" and "DEF" comes after "AB" and "DEF", hence the sprint right padded for each column up to the column's length (at least for all but the last column)

AlinChindea: > Also,
> could you return
> "string1#string2"
> using the delimeter "#"?

this doesn't seem to do anything. my version, below 👇 

`$workers = CentreUser::get()->sortBy(function() {
            return 'homeCentre.name#name';
        });`

AlinChindea: alas, found an easier solution. had to read an article I skipped the first time around...

AlinChindea: it's counterintuitive but if you have users collection and you want to sort by last name, first name, you'd think you can do `dd($users->sortBy('last_name')->sortBy('first_name'));`

actually, you need to put them in reverse order
`dd($users->sortBy('first_name')->sortBy('last_name'));`

👀 🕵️‍♂️

charlesstrange2: something like
```
$workers = CentreUser::get()->sortBy(function ($centreUser) {
  return $centreUser->homeCentre->name . "#" . $centreUser-name
}
```

app/Http/Controllers/Service/Admin/CentreUsersController.php
```diff
@@ -223,9 +223,7 @@ private function syncCentres($request, CentreUser $cu)
 
     public function download()
     {
-        $workers = CentreUser::get()->sortBy(function($centreUser) {
-            return sprintf('%-12s%s', $centreUser->homeCentre->name, $centreUser->name);
-        });
+        $workers = CentreUser::get()->sortBy('name')->sortBy('homeCentre.name');
```
charlesstrange2, line 226: as per https://github.com/laravel/ideas/issues/11#issuecomment-329866327 ?
give it a go.

AlinChindea: yes, it's similar to that but then there's another comment after [https://github.com/laravel/ideas/issues/11#issuecomment-329970458](https://github.com/laravel/ideas/issues/11#issuecomment-329970458) that says that the reverse sorting works accidentally and is based on undefined behaviour. I saw it it here, too [https://medium.com/@codebyjeff/the-keys-are-the-key-in-laravel-multiple-field-sortby-60cd5bf3f78b](https://medium.com/@codebyjeff/the-keys-are-the-key-in-laravel-multiple-field-sortby-60cd5bf3f78b)

so I don't know which one to choose... thoughts?

PR #415, opened 2020-06-03
5.5 to 5.6

tobybatch:


tests/Feature/Store/SearchPageTest.php
```diff
@@ -263,10 +263,11 @@ public function itPaginatesWhenRequired()
         ]);
 
         // Visit search page, make sure next page link is present and works
+        // <a class="page-link" href="http://arcv-store.test/registrations?page=2" rel="next" aria-label="Next »">›</a>
         $this->actingAs($centreUser, 'store')
             ->visit(URL::route('store.registration.index'))
-            ->see('<a href="' . URL::route('store.base') . '/registrations?page=2' . '" rel="next">»</a>')
-            ->click('»')
+            ->see('<a class="page-link" href="' . URL::route('store.base') . '/registrations?page=2' . '" rel="next" aria-label="Next »">›</a>')
```
katjam, line 269: Is this a change that was part of the upgrade?

tests/Unit/Passport/RoutesTest.php
```diff
@@ -56,6 +56,7 @@ public function testDontGetAccessTokenWithBadClientId()
 
         $this->assertEquals(json_decode($response, true), [
             'error' => 'invalid_client',
+            'error_description' => 'Client authentication failed',
```
katjam, line 59: Interesting, is the addition of description for client message flexibility?

tests/Feature/Store/SearchPageTest.php
tobybatch: Yes, well it broke the tests and I didn't change anything else.  It was the first thing I hit

tests/Unit/Passport/RoutesTest.php
tobybatch: Again, it just appeared.

tests/Feature/Store/SearchPageTest.php
katjam: It's the paginator here:
image

Probably a library that updated the default design to use `<` instead of `<<`.  Probably no one will notice but worth mentioning at least to @Nikomus to see if he thinks it significantly affects the UX.

I think we can override those in a lang file.

tests/Unit/Passport/RoutesTest.php
katjam: Worth looking at the Passport docs to note intent? I guess by default the field copies whatever we use in message. The change might mena our implementation isn't using a new feature/ improvement.

katjam: As your instinct goes though... we probably don;t want to get too stuck on anything trivial in terms of time / money spent. So your strategy is fair!

tests/Feature/Store/SearchPageTest.php
charlesstrange2, line 269: wondering if  `aria-label="Next »"` needs the `»`

tests/Unit/Passport/RoutesTest.php
charlesstrange2: agreed

tests/Feature/Store/SearchPageTest.php
Nikomus: It's probably clearer as one chevron as it's the next page link rather than a last page link.

Nikomus: It seems unhelpful that it's there. I guess this is the default that Illuminate\Pagination\LengthAwarePaginator is deciding should be there?

PR #418, opened 2020-06-05
1.10/1609 copy sweep

AlinChindea:
### Trello card 👉 [https://trello.com/c/96PsmTn6/1609-copy-sweep](https://trello.com/c/96PsmTn6/1609-copy-sweep)

### DIFF
--- a/app/Services/VoucherEvaluator/Evaluations/ChildIsSecondarySchoolAge.php
+++ b/app/Services/VoucherEvaluator/Evaluations/ChildIsSecondarySchoolAge.php
@@ -10,7 +10,7 @@ use Chalcedonyt\Specification\NotSpec;
 
 class ChildIsSecondarySchoolAge extends BaseChildEvaluation
 {
-    public $reason = 'is secondary school age they get no vouchers';
+    public $reason = 'is secondary school age they get';
     private $specification;
 
     /**
--- a/app/Services/VoucherEvaluator/Evaluations/ChildIsSecondarySchoolAge.php
+++ b/app/Services/VoucherEvaluator/Evaluations/ChildIsSecondarySchoolAge.php
@@ -10,7 +10,7 @@ use Chalcedonyt\Specification\NotSpec;
 
 class ChildIsSecondarySchoolAge extends BaseChildEvaluation
 {
-    public $reason = 'secondary school age';
+    public $reason = 'is secondary school age';
     private $specification;
 
     /**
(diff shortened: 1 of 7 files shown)


app/Services/VoucherEvaluator/Evaluations/FamilyHasNoEligibleChildren.php
```diff
@@ -11,7 +11,7 @@
 
 class FamilyHasNoEligibleChildren extends BaseFamilyEvaluation
 {
-    public $reason = 'in need of under ones to qualify primary schoolers';
+    public $reason = 'has no child under primary school age then children of primary school age get no vouchers';
```
charlesstrange2, line 14: These are bringing the voucher values into the rule text. 

The danger is that if we do that and circumtances change, then we have to rewrite this

the template might be done as 
"if {entity} {reason} {value} voucher(s)"

resources/views/store/edit_registration.blade.php
```diff
@@ -129,22 +129,22 @@ class="@if($errors->has('pri_carer'))invalid @endif" type="text"
...
                                 @endforeach
                             </ul>
                             @if(count($evaluations["disqualifiers"]) > 0)
-                                <p>[cCc] Don't credit :[cCc]</p>
+                                <p>Reminders:</p>
                                 <ul id="disqualifiers">
                                     @foreach($evaluations["disqualifiers"] as $disqualifier)
                                         <li>
-                                            [cCc]Where {{ strtolower(class_basename($disqualifier::SUBJECT)) }} is {{ $disqualifier->reason }}[cCc]
+                                            If {{ strtolower(class_basename($disqualifier::SUBJECT)) }} {{ $disqualifier->reason }}
```
charlesstrange2, line 147: This might need to change, see above

app/Services/VoucherEvaluator/Evaluations/ChildIsSecondarySchoolAge.php
```diff
@@ -10,7 +10,7 @@
 
 class ChildIsSecondarySchoolAge extends BaseChildEvaluation
 {
-    public $reason = 'secondary school age';
+    public $reason = 'is secondary school age they get no vouchers';
```
charlesstrange2, line 13: see below

AlinChindea:
> This will print "0" rather than "no" but I think is functionally good.

that's true but it doesn't sound bad. plus, we're more flexible if that rule changes... 
Screenshot 2020-06-05 at 15 55 07

charlesstrange2:
Forcing because travis / github is disconnected... again.

PR #420, opened 2020-06-08
1.10/1569 upgrade third party dependencies for service fonts ee

charlesstrange2:
Now with self hosted google fonts. 
The woff2 files are the ones we need, but i'm thinking we need the ttf's to live next to with the licences?

https://trello.com/c/PjEc8YmO/1569-upgrade-third-party-dependencies-for-service-ee


public/store/privacy_policy.html
```diff
@@ -1,6 +1,5 @@
 <meta charset="utf-8" />
-<link href="https://fonts.googleapis.com/css?family=Lato" rel="stylesheet">
-<link href="https://fonts.googleapis.com/css?family=Roboto+Slab" rel="stylesheet">
+<link rel="stylesheet" href="assets/google/fonts.css">
```
AlinChindea, line 2: why not be consistent throughout and use ` <link rel="stylesheet" href="{{ asset('store/assets/google/fonts.css') }}" >`?

PR #422, opened 2020-06-08
Add migration that runs in production mode only

charlesstrange2:
Ok, so this checks if, in prod only, a sponsor with SK exists, it then adds the evaluations required to modify the defaults.

https://trello.com/c/Ae1UC3WY/1594-write-migration-for-live-to-add-rules-for-sk

### DIFF
--- a/database/migrations/2020_06_08_140727_update__s_k_config_in_production.php
+++ b/database/migrations/2020_06_08_140727_update__s_k_config_in_production.php
@@ -1,0 +1,115 @@
+<?php
+
+use App\Evaluation;
+use App\Sponsor;
+use Illuminate\Database\Migrations\Migration;
+
+class UpdateSKConfigInProduction extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        // Only run in production
+        if (App::environment('production')) {
+            // Fetch the SK area
+            $sponsor = Sponsor::where('shortcode', 'SK')->first();
+            if (!$sponsor) {
+                // Log a warning
+                Log::warning(self::class . ": up() - unable to find SK area");
+                return;
+            }
+            
+            // Changes for crediting primary schoolers if they are qualified.
+            $modfyingEvaluations = [
+                // warn when primary schoolers are approaching end of school
+                new Evaluation([
+                    "name" => "ChildIsAlmostSecondarySchoolAge",
+                    "value" => "0",
+                    "purpose" => "notices",
+                    "entity" => "App\Child",
+                ]),
+                
+                // credit primary schoolers
+                new Evaluation([
+                    "name" => "ChildIsPrimarySchoolAge",
+                    "value" => "3",
+                    "purpose" => "credits",
+                    "entity" => "App\Child",
+                ]),
+                
+                // don't disqualify primary schoolers
+                new Evaluation([
+                    "name" => "ChildIsPrimarySchoolAge",
+                    "value" => null,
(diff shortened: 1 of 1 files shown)


database/migrations/2020_06_08_140727_update__s_k_config_in_production.php
```diff
@@ -0,0 +1,115 @@
...
+class UpdateSKConfigInProduction extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        // Only run in production
```
katjam, line 16: This will also run on staging, right? Since it is env production.

database/migrations/2020_06_08_140727_update__s_k_config_in_production.php
```diff
@@ -0,0 +1,115 @@
...
+        if (App::environment('production')) {
+            // Fetch the SK area
+            $sponsor = Sponsor::where('shortcode', 'SK')->first();
+            if (!$sponsor) {
```
katjam, line 20: On staging no sponsor will be found. Check logs to verify this? As a test?

database/migrations/2020_06_08_140727_update__s_k_config_in_production.php
```diff
@@ -0,0 +1,115 @@
...
+                    "purpose" => "disqualifiers",
+                    "entity" => "App\Family",
+                ]),
+            ];
+
+            // Add those to the sponsor
+            $sponsor->evaluations()->saveMany($modfyingEvaluations);
+
+            // Check operation
+            if ($sponsor->evaluations()->count() < count($modfyingEvaluations)) {
```
katjam, line 73: Could be more robust test, but this is safe enough as belt and braces.

database/migrations/2020_06_08_140727_update__s_k_config_in_production.php
```diff
@@ -0,0 +1,115 @@
...
+            // Fetch the SK area
+            $sponsor = Sponsor::where('shortcode', 'SK')->first();
+            if (!$sponsor) {
+                // Log a warning
+                Log::warning(self::class . ": down() - unable to find SK area");
+                return;
+            }
+
+            // Remove those evaluations
+            $sponsor->evaluations()->delete();
```
katjam, line 102: This will remove all evaluations - but since we know there were none at this point, is moot.

katjam:
Re. staging? Do we actually want the seeds to set an SK for the test cases? Is there already one?

database/migrations/2020_06_08_140727_update__s_k_config_in_production.php
charlesstrange2: yes. we should see warnings about no SK

database/migrations/2020_06_08_140727_update__s_k_config_in_production.php
charlesstrange2: yep

database/migrations/2020_06_08_140727_update__s_k_config_in_production.php
charlesstrange2: yep

charlesstrange2:
> Re. staging? Do we actually want the seeds to set an SK for the test cases? Is there already one?

There isn't one at the moment.

charlesstrange2:
Forcing, because, while travis passes, he isn't telling github.

Again.

PR #432, opened 2020-12-15
1.2/1647 bug cant create a delivery if weve never delivered vvv ee

charlesstrange2:
Looks like I needed a separate set of variables in my SQL and to be more explicit with sorting the output.

https://trello.com/c/8LvwcWSa/1647-bug-cant-create-a-delivery-if-weve-never-delivered-vvv-ee


AlinChindea (approved):
looks good from my end but I can't comment on the sql bits. 🐙

PR #439, opened 2020-12-22
Fix navbar link and route

charlesstrange2:
https://trello.com/c/SFf3qBEt/1683-bug-route-not-found-for-data-seeder


AlinChindea (approved):
🎅  approves!

PR #441, opened 2021-01-04
Fix group by

charlesstrange2:
First, it needs to be a "having", rather than a "where", when processing the computed value;

Secondly, SQLite and MySQL treat "having" differently
- SQLite likes to have everything grouped
- MySQL doesn't care

So we put a `groupBy('<some uid in the query>')` to keep SQLite happy


AlinChindea (approved):
🍑

PR #442, opened 2021-01-07
scramble the email unqiuely

charlesstrange2:
https://trello.com/c/Tih29A0V/1694-bug-unable-to-re-add-a-trader

### DIFF
--- a/app/Http/Controllers/Service/Admin/TradersController.php
+++ b/app/Http/Controllers/Service/Admin/TradersController.php
@@ -231,7 +231,7 @@ class TradersController extends Controller
                 $orphanUsers->each(function ($orphanUser) {
                     // make a Uuid for the email, which has a unique constraint
                     $hash = Uuid::uuid4()->toString();
-                    // localhost - it gets emailed, somehow, it'll be looped back to the box.
+                    // localhost - if it gets emailed, somehow, it'll be looped back to the box.
                     $orphanUser->update(['email' => $hash . '@localhost']);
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Service/Admin/TradersController.php
```diff
@@ -224,12 +225,17 @@ public function update(AdminNewUpdateTraderRequest $request, int $id)
...
-                User::whereIn('id', $orphanUsers)->delete();
-
+                    ->get();
+
+                // this is a slow iteration, but we don't expect too many users/trader
+                $orphanUsers->each(function ($orphanUser) {
+                    // make a Uuid for the email, which has a unique constraint
+                    $hash = Uuid::uuid4()->toString();
+                    // localhost - it gets emailed, somehow, it'll be looped back to the box.
+                    $orphanUser->update(['email' => $hash . '@localhost']);
```
AlinChindea, line 235: just curious about the `@localhost` bit 🌵

AlinChindea (approved):
🍫

app/Http/Controllers/Service/Admin/TradersController.php
charlesstrange2: I'm inventing a "valid", unique but undeliverable email address.

An email which is whatever@localhost *should* be routed to the current box. 
localhost is a reserved domain, and the internet won't deliver it.

It's arguable that we could just have stuck anything in the email field, but i'd prefer if it was at lease emaily.

PR #443, opened 2021-01-07
fix the sendResetResponse method parameters which have changed in an …

charlesstrange2:
So, it seems the authentication system has changed a method signature we override.
This fixes that.


AlinChindea (approved):
🛳️

PR #445, opened 2021-01-11
1.12/1662 traders download sheet with formatted dates

AlinChindea:
Trello 🎫 [https://trello.com/c/92Q7iwBJ/1662-a-means-to-download-the-trader-view-as-a-spreadsheet-v-ee](https://trello.com/c/92Q7iwBJ/1662-a-means-to-download-the-trader-view-as-a-spreadsheet-v-ee)

### DIFF
--- a/app/Http/Controllers/Service/Admin/TradersController.php
+++ b/app/Http/Controllers/Service/Admin/TradersController.php
@@ -210,7 +210,7 @@ class TradersController extends Controller
                                         'revoked' => true,
                                     ]);
 
-                                Log::info('removing token for user '. $user->id);
+                                Log::info('removing token for user ' . $user->id);
                                 $accessToken->revoke();
                             } else {
                                 Log::info('no tokens to revoke');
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Service/Admin/TradersController.php
```diff
@@ -259,6 +259,8 @@ public function download()
         $traders = $this->getTraders();
         $csvExporter = new Export();
         $csvExporter->beforeEach(function ($trader) {
+            $trader->joined = $trader->created_at->format('d-m-Y');
+            $trader->disabled = $trader->disabled_at ? $trader->disabled_at->format('d-m-Y') : $trader->disabled_at;
```
charlesstrange2, line 263: Is the behaviour of "null" to print nothing in the cell?

charlesstrange2 (approved):
okedoke 👍

app/Http/Controllers/Service/Admin/TradersController.php
AlinChindea: yes, `null` prints nothing in the cell. also, 'null' just explodes the whole thing - there's no method `format` on `null` - hence the ternary.

PR #447, opened 2021-01-13
1.12/1686 bug error responses not being returned to bad requests

charlesstrange2:
Updated some blades to add the error boxes
added some tests

https://trello.com/c/eHWdtqcR/1686-bug-error-responses-not-being-returned-to-bad-requests

### DIFF
--- a/app/Http/Requests/AdminNewUpdateMarketRequest.php
+++ b/app/Http/Requests/AdminNewUpdateMarketRequest.php
@@ -3,6 +3,7 @@
 namespace App\Http\Requests;
 
 use Illuminate\Foundation\Http\FormRequest;
+use Illuminate\Validation\Rule;
 
 class AdminNewUpdateMarketRequest extends FormRequest
 {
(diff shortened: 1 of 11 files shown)


app/Http/Requests/AdminNewUpdateMarketRequest.php
```diff
@@ -24,9 +25,21 @@ public function authorize()
      */
     public function rules()
     {
+        // Get the ID off the route
+        $id = $this->route('id') ?? null;
+        $sponsor_id = $this->only('sponsor');
```
AlinChindea, line 30: does the only method plucks the id of whatever var name we're passing in there? nice

resources/views/service/traders/edit.blade.php
```diff
@@ -72,7 +72,7 @@ class="styled-checkbox {{ $errors->has('disabled') ? 'error' : '' }}"
                     </div>
                 </div>
                 @include('service.partials.addTraderUser', ['users' => $trader->users ])
-                <button type="submit" id="updateMarket">Update All</button>
+                <button type="submit" id="updateTrader">Update All</button>
```
AlinChindea, line 75: oops 😄

app/Http/Requests/AdminNewUpdateMarketRequest.php
AlinChindea: if it's the Laravel collections method only - https://laravel.com/docs/6.x/collections#method-only - i think your argument here should be `sponsor_id` if you need the id of the sponsor. same for market in the AdminNewUpdateTraderRequest

charlesstrange2: `$this` refers to the formRequest, which is a type of Request. only, will pull the input off. 
However, you may be right, i don't think i want an array, which only returns.

AlinChindea (approved):
🐙 okey dokey

PR #448, opened 2021-01-14
1.12/1705 bug css on edit trader pages not the right selector

AlinChindea:
#### Trello 🎫  [https://trello.com/c/eVuEWYo8/1705-bug-css-on-edit-trader-page-is-not-targetting-the-right-element](https://trello.com/c/eVuEWYo8/1705-bug-css-on-edit-trader-page-is-not-targetting-the-right-element)

### DIFF
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -225,7 +225,7 @@ h1 {
       background-color: lighten($arc_rose, 15%);
       color: $arc_white;
     }
-    &#updateMarket {
+    &.updateTrader {
       margin-top: 2em;
     }
     &#add-user {
--- a/resources/views/service/traders/create.blade.php
+++ b/resources/views/service/traders/create.blade.php
@@ -54,7 +54,7 @@
 
                 @include('service.partials.addTraderUser')
 
-                <button type="submit" id="updateTrader">Save Trader</button>
+                <button type="submit" class="updateTrader">Save Trader</button>
             </form>
         </div>
     </div>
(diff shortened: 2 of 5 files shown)


resources/assets/sass/app.scss
```diff
@@ -225,7 +225,7 @@ h1 {
       background-color: lighten($arc_rose, 15%);
       color: $arc_white;
     }
-    &#updateMarket {
+    &.updateTrader {
```
charlesstrange2, line 228: Does this affect the market forms which might use updateMarket on thier buttons?

https://github.com/neontribe/ARCVService/blob/1.12/release/resources/views/service/markets/create.blade.php
https://github.com/neontribe/ARCVService/blob/1.12/release/resources/views/service/markets/edit.blade.php

charlesstrange2 (commented):
Changing the css to a class and altering its name might affect other uses?

resources/assets/sass/app.scss
AlinChindea: no, those are nicely spaced.

charlesstrange2:
Ok, add a check  for them in the ACs

resources/assets/sass/app.scss
AlinChindea: the original css property was not intended to target those in the makert create and edit pages. it was just change of the selector that lead to that. but there's no need, those buttons are nicely spaced, not overlapping with any other elements event at smaller screen sizes.

charlesstrange2 (approved):
👍

PR #463, opened 2021-09-23
Adding date parameter to only get last 6 months of vouchers

karalan:
I guess we need to test this on staging but think it should do the trick?


karalan:
@charlesstrange2 updated to match hotfix implementation

charlesstrange2:
closing as _should_ be implemented.

PR #465, opened 2021-10-04
1.12 hotfix 732 move work into database

charlesstrange2:
@karalan and @charlesstrange2 had a crack at making things faster and not broken.


charlesstrange2:
closed per new ticket

PR #475, opened 2022-02-21
1.12/1749 admin store package updates vvv e

neonleann:
https://trello.com/c/e4yunEIS/1749-admin-store-package-updates-vvv-e

I have done what I can. The packages updated and the ones I could do are on the ticket.
Regarding the security notices, I added resolutions for all the ones I could. The ones that couldn't be updated are:
is-svg - although yarn list says nothing is using it
ansi-html - there's no fix
postcss - I think everything is using that
node-forge - laravel-mix has issues with an update
nth-check - same as above
glob-parent - same as above

All the tests (except for the two I mentioned) pass. Yarn audit is down to Severity: 2 Low | 13 Moderate | 4 High. All listed vulnerabilities are the ones up there.


yarn-error.log
```diff
@@ -0,0 +1,6990 @@
+Arguments:
```
charlesstrange2: Add this file to gitignore to avoid committing

package.json
```diff
@@ -11,13 +11,25 @@
...
+    "jquery": "^3.6.0",
+    "jquery-ui": "^1.13.1",
     "laravel-mix": "^5.0.9",
-    "resolve-url-loader": "^3.1.2",
-    "sass": "^1.23.7",
+    "resolve-url-loader": "^5.0.0",
+    "sass": "^1.49.8",
     "sass-loader": "^7.3.1",
-    "vue-template-compiler": "^2.6.10"
+    "vue-template-compiler": "^2.6.14"
```
charlesstrange2, line 21: Do we use vue.js in this?
see if you can find any references to it.

charlesstrange2 (changes requested):

package.json
neonleann: This "Vue.component("example",n(34));new Vue({el:"#app"})}" is in public/store/js/app.js, if that counts.

charlesstrange2: interesting... i'll make a research item to determine if we can pull it out at another time!

charlesstrange2 (approved):
Okedoke, 👍

PR #479, opened 2022-02-23
Nice message about collection date.

neonleann:
https://trello.com/c/dqwILa4K/1739-reduce-the-ability-to-forward-date-voucher-bundles-in-store-vv-ee

Note: after trying it out, I don't think I can write the tests. It won't press the button because it's not a link or in a form. Plus, even if I could get it to the correct page, the error is controlled by JS, which I thought we couldn't run tests for. If I'm wrong, I will happily add in the tests. For now, I removed it from the plan.

### DIFF
--- a/resources/views/store/manage_vouchers.blade.php
+++ b/resources/views/store/manage_vouchers.blade.php
@@ -171,6 +171,7 @@
                             <i class="fa fa-calendar"></i>
                             <div>
                                 <label for="collected-on">Collected on:</label>
+                                <div id="dateError" style="display:none;"></div>
                                 <input id="collected-on" name="collected_on" value="<?php echo date('Y-m-d');?>" type="date">
                             </div>
                         </div>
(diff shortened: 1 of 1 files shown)


resources/views/store/manage_vouchers.blade.php
```diff
@@ -270,6 +271,21 @@ function () {
...
+                  var chosenDateJS = new Date(chosenDate);
+                  var sixWeeks = new Date();
+                  sixWeeks.setDate(sixWeeks.getDate() + 42);
+                  if (chosenDateJS > sixWeeks) {
+                    $('#dateError').text('Please choose a date within six weeks of today');
+                    $('#dateError').css('font-size', '12px');
+                    $('#dateError').css('color', 'red');
+                    $('#dateError').show();
+                  } else {
+                    $('#dateError').hide();
```
charlesstrange2, line 286: `display:none` is different to `display:hidden`; does `hide()` actually do `display:none`?

charlesstrange2 (approved):
ok

PR #484, opened 2022-02-28
1.13/1734 improvements to childrens centre change names v ee

neonleann:
https://trello.com/c/B8WopCQB/1734-improvements-to-childrens-centre-change-names-v-ee
https://trello.com/c/zOi490Jq/1735-improvements-to-childrens-centre-sort-by-area-v-ee

### DIFF
--- a/app/Http/Controllers/Service/Admin/CentresController.php
+++ b/app/Http/Controllers/Service/Admin/CentresController.php
@@ -5,6 +5,7 @@ namespace App\Http\Controllers\Service\Admin;
 use App\Centre;
 use App\Http\Controllers\Controller;
 use App\Http\Requests\AdminNewCentreRequest;
+use App\Http\Requests\AdminUpdateCentreRequest;
 use Auth;
 use DB;
 use Exception;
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Service/Admin/CentresController.php
```diff
@@ -99,4 +100,47 @@ public function store(AdminNewCentreRequest $request)
...
     }
+
+    /**
+     * Show the form for editing a Centre.
+     *
+     * @return Factory|View
+     */
+    public function edit($id)
+    {
+        $centre = Centre::where('id', $id)->first();
```
charlesstrange2, line 111: laravel has a shortcut for this as Centre::find($id)

[ As well as findOrFail(), which I see you use below to force an exception if it cant find one there's findOrCreate($id, [array of attributes]) which will _create_ one if it can't find it ]

app/Http/Controllers/Service/Admin/CentresController.php
```diff
@@ -99,4 +100,47 @@ public function store(AdminNewCentreRequest $request)
...
+        return view('service.centres.edit', compact('centre'));
+    }
+
+    /**
+     * Show the form for editing a Centre's name.
+     *
+     * @return Factory|View
+     */
+    public function update(AdminUpdateCentreRequest $request)
```
charlesstrange2, line 120: Laravel magic can get the dependency injector to pull the id out of the route parameters IF we send id as part of the URL  by doing `update(AdminUpdateCentreRequest $request, integer $id)`

if you type hint  `update(AdminUpdateCentreRequest $request, Centre $id)` it will _also_ do an automagical fetch of the model from the database. _however_ in this case it looks like you want to do that by hand with findOrFail so we can manually do something with the error?

routes/service.php
```diff
@@ -102,6 +102,14 @@
         'as' => 'admin.centres.store',
         'uses' => 'Admin\CentresController@store',
     ]);
+    Route::post('centres/update', [
```
charlesstrange2, line 105: Semantically, a PUT or PATCH to /centres with a single item is an update of the centres collection
OR
a PUT to  `centres/{id}/update`, would be more explicit, so we know which ID we're updating without having to grab it from the input (See above).

charlesstrange2 (changes requested):
consider route changes?

app/Http/Controllers/Service/Admin/CentresController.php
neonleann: Changed.

app/Http/Controllers/Service/Admin/CentresController.php
neonleann: It took me a while to work it out, but this is changed.

routes/service.php
neonleann: Updated as requested.

charlesstrange2 (approved):
okedoke! 🦜

PR #486, opened 2022-03-01
1.13/1651 trader has 6mb payment history which refuses to load in app vvv eee

charlesstrange2:
This
- force holds a package "symfony/console": "4.4.36", becuase it stops tinker working at 4.4.37
- optimises TraderController to use eloquent methods that Sqlite likes so we can test.
- Pagiantion to 15/page (the default)
- links for first, next, prev ,last (and current) in the headers
https://trello.com/c/UqyVMi3p/1651-trader-has-6mb-payment-history-which-refuses-to-load-in-app-vvv-eee


neonleann (approved):
Everything certainly worked fine when I was using it to test my seeds. Great work 💳.

PR #488, opened 2022-03-07
fix CORS to it will show links

charlesstrange2:
Second pull request to fix the CORS system to permit the `Links` header


neonleann (approved):
Short and sweet, but if it works, it works 🔗.

PR #497, opened 2022-03-09
1.13/1764 design changes

Nikomus:
https://trello.com/c/BfOlDp2u/1764-design-changes-v-ee

Individual changes listed per commit

### DIFF
--- a/resources/views/store/edit_registration.blade.php
+++ b/resources/views/store/edit_registration.blade.php
@@ -101,7 +101,7 @@
                 </div>
             </div>
 
-            <div class="col collect short-height">
+            <div class="col collect fit-height">
                 <div>
                     <img src="{{ asset('store/assets/info-light.svg') }}" alt="logo">
                     <h2>This family</h2>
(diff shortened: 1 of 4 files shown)


resources/views/store/edit_registration.blade.php
```diff
@@ -180,47 +180,43 @@ class="@if($errors->has('pri_carer'))invalid @endif" type="text"
...
                         <i class="fa fa-ticket button-icon" aria-hidden="true"></i>Go to voucher manager
                     </div>
                 </a>
-            </div>
-        </form>
+                
+                @if (!isset($registration->family->leaving_on) )
 
-        @if (!isset($registration->family->leaving_on) )
-            <form action="{{ URL::route('store.registration.family',['registration' => $registration]) }}" method="post"
```
charlesstrange2, line 197: This form hit a different endpoint?

charlesstrange2 (changes requested):
check the old leaving form - it looks like a different endpoint

resources/views/store/edit_registration.blade.php
Nikomus: I think my formaction went missing with some markup shuffling

charlesstrange2: Ah! I see you've merged the two forms and then got the second button to submit all the formdata to the update route too?

charlesstrange2: is that why we do Ah! Is that why formNoValidate?

charlesstrange2: hmm. this ticket says we support ie11 😨 which formNoValidate isn't a thing for 😬 

Should we have the two forms back?

charlesstrange2 (changes requested):
reinstate 2 forms?

charlesstrange2 (approved):
okedoke :+1:

PR #500, opened 2022-03-11
1.13/1742 a child in scotland can be deferred vvv ee

neonleann:
https://trello.com/c/LyLz9dv4/1742-a-child-in-scotland-can-be-deferred-vvv-ee


app/Child.php
```diff
@@ -137,4 +149,33 @@ public function family()
...
     }
+
+    /**
+     * Get info about whether child can defer
+     *
+     * @return boolean
+     */
+    public function getCanDeferAttribute()
+    {
+      $notices = $this->family->registrations[0]->getEvaluator()->evaluations["App\Child"]["notices"];
```
charlesstrange2, line 160: 1. This is quite a deep dependent call that assumes the Child class knows a lot about the arrangement of the Family and registrations (cf. "encapsulation"); 

amend this->getEvaluator() as above and then call it here. This method doesn't need  to be the one that does that work. better to `this->getEvaluator()` where that function would check for a family and grab it's evaluator if necessary. 

2. Then once we have a rules set, check that `->evaluations["App\Child"]["notices] actually exists, else you risk a PHP 500 error when it tries to pull a key off an array. 

could do it with `->evaluations["App\Child"]["notices"] ?? []` as i think we want an empty array if it _can't_ find notices that apply to a Child.

app/Child.php
```diff
@@ -137,4 +149,33 @@ public function family()
     {
         return $this->belongsTo('App\Family');
     }
+
+    /**
+     * Get info about whether child can defer
+     *
+     * @return boolean
+     */
+    public function getCanDeferAttribute()
```
charlesstrange2, line 158: I think the indent is one-off on this function.

app/Child.php
```diff
@@ -137,4 +149,33 @@ public function family()
...
+              "value" => 0,
+              "purpose" => "notices",
+              "entity" => "App\Child",
+        ])
+      ];
+      $rulesMods = collect($rule);
+      $evaluator = EvaluatorFactory::make($rulesMods);
+      $evaluation = $evaluator->evaluate($this);
+      $notices = $evaluation["notices"];
+      if (count($notices)) {
```
charlesstrange2, line 176: Probably should check of `count($notices) > 0`
or ternary to 

`return (count($notices) > 0) ? true : false;`

but could also be cast to a bool

`return (bool) count($notices);`

charlesstrange2: Also, check that the notice is explicity the ScottishChildCanDefer one
Otherwise, we assume that the default notices the EvaluatorFactory holds will never be altered to inlude _other_ notices about a child

app/Child.php
```diff
@@ -137,4 +149,33 @@ public function family()
...
+     *
+     * @return boolean
+     */
+    public function getCanDeferAttribute()
+    {
+      $notices = $this->family->registrations[0]->getEvaluator()->evaluations["App\Child"]["notices"];
+      if (!array_key_exists('ScottishChildCanDefer', $notices)) {
+        return false;
+      }
+      $rule = [
```
charlesstrange2, line 164: Here we're saying
IF the rules you would evauate this Child don't include the ScottishChildCanDefer,
THEN return false; this child cannot be tested for deferring. 

If they _do_ have ScottishChildCanDefer, then make a new validator that _only_ validates the child against the ScottishChildCanDefer rule.

is this redundant?

app/Http/Controllers/Store/RegistrationController.php
```diff
@@ -195,6 +195,18 @@ public function edit(Registration $registration)
         } else {
             abort(404, 'Registration not found.');
         }
+        $evaluations = $registration->centre->sponsor->evaluations;
```
charlesstrange2, line 198: This gets the rules we'd append to the defaults

could also be `$registration->getEvaluator()->evaluations` IF it's ok to operate with the combined rules.

app/Http/Controllers/Store/RegistrationController.php
```diff
@@ -195,6 +195,18 @@ public function edit(Registration $registration)
...
+        $is_scottish = false;
+        foreach ($evaluations as $key => $evaluation) {
+          if ($evaluation['name'] == 'ScottishChildCanDefer') {
+            $is_scottish = true;
+          }
```
charlesstrange2, line 203: we can put `break;` here to exit a foreach loop early, when we find what we want.

app/Services/VoucherEvaluator/Evaluations/ScottishChildCanDefer.php
```diff
@@ -30,7 +30,7 @@ public function test($candidate)
         $monthNow = Carbon::now()->month;
         $schoolStartMonth = config('arc.scottish_school_month');
         // Check we're in the start month or the one before.
-        if ($schoolStartMonth - $monthNow > 1) {
+        if (($schoolStartMonth - $monthNow > 1) || ($schoolStartMonth - $monthNow < 0)) {
```
charlesstrange2, line 33: resolves to success when:
- a month before (1)
- the month (0)

and fail when _after_ the date OR before 1-month-before

👍 that's my understanding

app/Services/VoucherEvaluator/Evaluations/ScottishChildIsBetweenOneAndPrimarySchoolAge.php
```diff
@@ -0,0 +1,78 @@
...
+        return true;
+      }
+      if ($year < 4) {
+        return false;
+      }
+      // Otherwise, check if the start month has gone.
+      if ($schoolStartMonth - $monthNow < 0) {
+        $isAtSchool = false;
+        // Are they still between 4 1 and 4 11 and not deferred OR are they over 5?
+        if (((($year == 4 && $month >=1) || $year < 5) && !$candidate->deferred) || $year >= 5) {
```
charlesstrange2, line 69: probably should be === rather than ==
this appears in some other files above too.

app/Services/VoucherEvaluator/Evaluations/ScottishChildIsPrimarySchoolAge.php
```diff
@@ -0,0 +1,64 @@
...
+          return $this->success();
+        }
+        if ($year < 4) {
+          return $this->fail();
+        }
+        // Otherwise, check if the start month has gone.
+        if ($schoolStartMonth - $monthNow < 0) {
+          $isAtSchool = false;
+          // Are they still between 4 1 and 4 11 and not deferred OR are they over 5?
+          if (((($year == 4 && $month >=1) || $year < 5) && !$candidate->deferred) || $year >= 5) {
```
charlesstrange2, line 52: `===`

app/Http/Controllers/Store/RegistrationController.php
```diff
@@ -218,6 +230,8 @@ public function edit(Registration $registration)
                 'entitlement' => $valuation->getEntitlement(),
                 'verifying' => $registration->getEvaluator()->isVerifyingChildren(),
                 'evaluations' => $evaluations,
+                'is_scottish' => $is_scottish,
```
charlesstrange2, line 233: perhaps "can_defer"?

resources/views/store/edit_registration.blade.php
```diff
@@ -70,6 +70,9 @@ class="@if($errors->has('pri_carer'))invalid @endif" type="text"
                             @if ( $verifying )
                             <td class="verified-col">ID</td>
                             @endif
+                            @if ( $is_scottish )
```
charlesstrange2, line 73: possibly "can_defer"?

charlesstrange2 (changes requested):

app/Child.php
neonleann: Done.

app/Services/VoucherEvaluator/Evaluations/ScottishChildCanDefer.php
neonleann: I believe so.

app/Services/VoucherEvaluator/Evaluations/ScottishChildIsBetweenOneAndPrimarySchoolAge.php
neonleann: All done.

app/Services/VoucherEvaluator/Evaluations/ScottishChildIsPrimarySchoolAge.php
neonleann: Done.

resources/views/store/edit_registration.blade.php
neonleann: I've gone with $deferrable, since I've used can_defer elsewhere.

PR #506, opened 2022-03-17
Fixing an issue with create registration.

neonleann:
https://trello.com/c/FZXpi1Bi/1791-default-dropdown-for-new-family-is-applying-and-prompts-are-not-displaying-when-family-details-are-first-saved


charlesstrange2 (approved):
okedoke!

PR #508, opened 2022-03-24
1.13/1798 hotfix history performance

charlesstrange2:
Changes to speed up things a little.
https://trello.com/c/GC94NpEi/1798-hotfix-history-performance

Turns out inner joins in the context are much faster than left joins, as inner joins don't bring back a bunch of null columns
https://www.sqlshack.com/learn-sql-inner-join-vs-left-join/


neonleann (approved):
Lovely stuff 🤞.

PR #509, opened 2022-03-24
1.13/1799-hotfix-replace-buttons-to-print

charlesstrange2:
https://trello.com/c/3i84b8Nt/1799-hotfix-replace-buttons-to-print
uncomment button

apparently, the DO need that button! we should put that back.


neonleann (approved):
Welcome back button! 👋

PR #515, opened 2022-04-14
Filtering on Reimbursed Date instead (Was - A nice simple way to split the file into two).

neonleann:
https://trello.com/c/QB9oJR9K/758-hotix-mvl-issue


app/Console/Commands/CreateMasterVoucherLogReport.php
```diff
@@ -218,17 +218,12 @@ public function handle()
...
-                      ) {
-                    unset($chunk[$k]);
-                    continue;
-                  }
-                  if (!is_null($voucher['Date Issued'])
-                        && (strtotime($voucher['Date Issued']) < strtotime('2020-01-01'))
-                      ) {
-                    unset($chunk[$k]);
-                    continue;
+                  if (!is_null($voucher['Reimbursed Date'])) {
```
charlesstrange2, line 221: by here we've 

- removed all the "printed" ones that trigger `containsOnlyNull()`
- removed anything that has a voided date (can that affect a voucher that was reimbursed?)

then, if we have a reimbursed date
- we check it's not to old
- and drop it if it is.

--- 

so we should end up with
no printed, no voided and no _old_ reimbursed
but new reimbursed and ... everything else?

neonleann: That sounds about right (no vouchers that are **only** printed).
I'm pretty certain that Michelle agreed that anything with a void date should go.

We should end up with vouchers that have been reimbursed AFTER April 1st 2021 and any that have not been reimbursed but HAVE been at least dispatched.

charlesstrange2 (approved):
Awesome, let's hack it into place

PR #522, opened 2022-05-23
1.14/1853 hotfix set cantap to default false

charlesstrange2:
https://trello.com/c/W6RHkcrL/1853-hotfix-set-cantap-to-default-false


tests/Unit/Controllers/Service/Admin/SponsorControllerTest.php
```diff
@@ -130,10 +130,12 @@ public function testICanStoreASponsor()
             ->see($data["voucher_prefix"])
         ;
 
-        // Find the Sponsor
+        // Find the Sponsor, and check it's attributes
```
neonleann, line 133: I do hope I wasn't responsible for that erroneous apostrophe.

neonleann (approved):
Yes. This is good (apart from the apostrotastrophe).

PR #523, opened 2022-05-23
Amend tests to remove is

lauragurney:
Tests were failing because I did not update them. This commit fixes that.


charlesstrange2 (approved):
Okedoke!

PR #524, opened 2022-05-23
1.14/1844 a centre can determine its mode vvv ee

charlesstrange2:
After this is done, on fetch

`Sponsor::programme` should return an INT
`Sponsor::programme_name` will do a look up on the arc config files and return a name/key

We can navigate to the Sponsor by `Auth::user()->centre->sponsor` if we need to pluck that off to determine a switch. 

At the moment that's fixed. since we don't need to carry much information about Social Prescribing as a _thing_ at this point, a name lookup is sufficient. Should we expand that to needing more properties, we can create a table at a later date.

https://trello.com/c/szywQIcL/1844-a-centre-can-determine-its-mode-vvv-ee

### DIFF
--- a/config/arc.php
+++ b/config/arc.php
@@ -98,7 +98,7 @@ return [
     |--------------------------------------------------------------------------
      */
     'programmes' => [
-        'standard',
-        'social prescribing',
+        'Standard',
+        'Social Prescribing',
     ],
 ];
--- a/config/arc.php
+++ b/config/arc.php
@@ -91,4 +91,14 @@ return [
     */
     'first_delivery_date' => env('ARC_FIRST_DELIVERY_DATE', '2019-09-26'),
 
+    /*
+    |--------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


config/arc.php
```diff
@@ -91,4 +91,14 @@
...
     'first_delivery_date' => env('ARC_FIRST_DELIVERY_DATE', '2019-09-26'),
 
+    /*
+    |--------------------------------------------------------------------------
+    | the programme names we are running
+    | eg. "standard", "social prescribing"
+    |--------------------------------------------------------------------------
+     */
+    'programmes' => [
+        'standard',
```
neonleann, line 101: Can we have the words starting with capitals? It looks very strange otherwise (imo).

neonleann (approved):
Other than my aversion to programme names starting with lower case letters, everything looks fine and seems to work a treat 🔠.

lauragurney (approved):
Seems to be working for me!

PR #525, opened 2022-05-24
1849 remove reasons and notices for SP

lauragurney:
Notices such as "A child is almost primary school age" should not appear for SP in voucher-manager and edit family.
Bullet point "Has 6 children registered (more)" should not not appear for SP in edit family.

$programme has been added for RegistrationController and BundleController so this can be checked in the blades before displaying the correct option. $programme [0] is Standard. 

https://trello.com/c/DZkUtpFd/1849-as-an-sp-user-i-do-not-need-to-see-the-reasons-why-i-am-getting-these-vouchers-vv-e


neonleann (approved):
Looks good!

PR #527, opened 2022-05-25
1.14/1845 entities respond correctly with their name depending on mode vvv ee

charlesstrange2:
A trait that can be applied to a model to provide it with a static getAlias, so long as the various constants are configured correctly.

also a test of the trait which gets a little wierd making some example models on the fly...

https://trello.com/c/OT8PQuSE/1845-entities-respond-correctly-with-their-name-depending-on-mode-vvv-ee

### DIFF
--- a/app/Carer.php
+++ b/app/Carer.php
@@ -2,10 +2,23 @@
 
 namespace App;
 
+use App\Traits\Aliasable;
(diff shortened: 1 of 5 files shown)


lauragurney (approved):
I like the look of this.

PR #529, opened 2022-05-26
1.14/1855 remove droppers for SP

lauragurney:
This PR removes the droppers for SP users, plus changes the rules to "sometimes" instead of "required" which then allows null to be recorded in the DB when it's not present (which it won't ever be for SP)

Also changed tests to be able to pass with this missing as the check is no longer necessary.

Full details here: https://trello.com/c/uNKj35dv/1855-should-we-record-nrpf-and-hs-or-bs-for-social-prescribing-vvv-ee

### DIFF
--- a/app/Http/Controllers/Store/RegistrationController.php
+++ b/app/Http/Controllers/Store/RegistrationController.php
@@ -164,11 +164,13 @@ class RegistrationController extends Controller
         $sponsor = $user->centre->sponsor;
         $evaluator = EvaluatorFactory::make($sponsor->evaluations);
         $verifying = $evaluator->isVerifyingChildren();
+        $programme = Auth::user()->centre->sponsor->programme;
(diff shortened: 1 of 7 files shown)


neonleann (commented):
I may have misunderstood, but doesn't the "No recourse to public funds (NRPF) family?" dropper need to go as well? I only see references to the "Are you receiving Healthy Start or Best Start?" dropper.

lauragurney:
> I may have misunderstood, but doesn't the "No recourse to public funds (NRPF) family?" dropper need to go as well? I only see references to the "Are you receiving Healthy Start or Best Start?" dropper.

It has gone, where are you seeing it?

lauragurney:
Yeah I realised I didn't display that well in the code, I basically just extended the long if statement to end after the droppers

PR #530, opened 2022-05-26
Make the Centre name unique.

neonleann:
https://trello.com/c/MOvIDi5E/1854-bug-transient-seed-fail-for-registrations-prefix-dupe


lauragurney (approved):
Noice.

PR #536, opened 2022-05-30
1.14/1817 an sp area has a set of rules that make sense vvv eee

neonleann:
https://trello.com/c/FeXQ11Yp/1817-an-sp-area-has-a-set-of-rules-that-make-sense-vvv-eee

### DIFF
--- a/app/Services/VoucherEvaluator/Evaluations/DeductFromCarer.php
+++ b/app/Services/VoucherEvaluator/Evaluations/DeductFromCarer.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace App\Services\VoucherEvaluator\Evaluations;
+
+use Carbon\Carbon;
+
+class DeductFromCarer extends BaseChildEvaluation
+{
+    public $reason = 'DEDUCT';
+    private $specification;
+
+    /**
+     * DeductFromCarer constructor.
+     * @param int|null $value
+     */
+    public function __construct(Carbon $offsetDate = null, int $value = null)
+    {
+        parent::__construct($offsetDate, $value);
+    }
+
+    public function test($candidate)
+    {
+        parent::test($candidate);
+        $children = $candidate->family->children;
+        // This rule should only apply to the first 'child', who will be the carer.
+        return ($candidate->id === $children[0]->id)
+            ? $this->success()
+            : $this->fail()
+        ;
+    }
+}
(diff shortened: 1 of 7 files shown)


app/Services/VoucherEvaluator/Evaluations/DeductFromCarer.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        parent::__construct($offsetDate, $value);
+    }
+
+    public function test($candidate)
+    {
+        parent::test($candidate);
+        $children = $candidate->family->children;
+        // This rule should only apply to the first 'child', who will be the carer.
+        return ($candidate->id === $children[0]->id)
```
charlesstrange2, line 26: this checks if the carer ID is the same as the same as the first child's ID, which is almost certainly not the case. 
What this rule should do is check to see if the candidate (Family?) has at least one child.

we can
- rename the class to "FamilyHasOneOrMoreChildren"; this rule tests is a family has at least one child
- set the test to return `$candidate->has('children');`

When we program the DB with this rule, we apply a negative credit to it, and the system takes care of applying that when this rule is true.

app/Services/VoucherEvaluator/Evaluations/HouseholdMember.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace App\Services\VoucherEvaluator\Evaluations;
+
+use Carbon\Carbon;
+
+class HouseholdMember extends BaseChildEvaluation
+{
+    public $reason = 'is member of the household';
```
charlesstrange2, line 9: probably drop the `is` to avoid plural problems

app/Services/VoucherEvaluator/Evaluations/DeductFromCarer.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace App\Services\VoucherEvaluator\Evaluations;
+
+use Carbon\Carbon;
+
+class DeductFromCarer extends BaseChildEvaluation
+{
+    public $reason = '';
```
charlesstrange2, line 9: possibly `one or more Participants`

charlesstrange2 (changes requested):
condition for the family needing at least one child is dependent on IDs rather than having some. :-)

charlesstrange2 (approved):
okedoke

PR #537, opened 2022-05-30
1.14/1811 refactoring signup screens(VVV / EEE)

lauragurney:
https://trello.com/c/d0rhy4wu/1811-refactoring-signup-screens-vvv-eee

This is the "add new family/household" functionality. Each column has been separated into its own file for "normal" and SP scenarios. The JS has been stripped out and put in its own file in store/js (which might be the wrong place so do let me know)

Waiting for further work on the age field, so that will not save  yet, this is the skeleton of that work. 

Happy to demo if needed.


lauragurney:
Closing for now until everything is sorted

PR #538, opened 2022-05-31
1.14/1857 front end for sp area rules

neonleann:
https://trello.com/c/HMlDB6mz/1857-front-end-for-sp-area-rules

### DIFF
--- a/routes/service.php
+++ b/routes/service.php
@@ -138,6 +138,14 @@ Route::group(['middleware' => 'auth:admin'], function () {
         'as' => 'admin.sponsors.store',
         'uses' => 'Admin\SponsorsController@store',
     ]);
+    Route::get('sponsors/{id}/edit', [
+        'as' => 'admin.sponsors.edit',
+        'uses' => 'Admin\SponsorsController@edit',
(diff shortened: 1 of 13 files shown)


routes/service.php
```diff
@@ -138,6 +138,14 @@
         'as' => 'admin.sponsors.store',
         'uses' => 'Admin\SponsorsController@store',
     ]);
+    Route::get('sponsors/{id}/edit', [
+        'as' => 'admin.sponsors.edit',
+        'uses' => 'Admin\SponsorsController@edit',
+    ])->where('id', '^[0-9]+$');
+    Route::put('sponsors/{id}/update', [
```
charlesstrange2, line 145: this can just be put 'sponsors/{id}'

charlesstrange2 (changes requested):
Knock the "update" off the put route :-)

routes/service.php
neonleann: It is gone, as so is the update one.

charlesstrange2 (approved):
:heavy_check_mark:

PR #540, opened 2022-06-01
1.14/1858 bug carbon behaves weirdly on the 31st may

neonleann:
https://trello.com/c/3fANuYs9/1858-bug-carbon-behaves-weirdly-on-the-31st-may


charlesstrange2 (approved):
Okedoke :spiral_calendar: 

Further work: Does this happen in other places?

PR #541, opened 2022-06-01
1.14/1811 refactoring signup and edit reg screens

lauragurney:
The voucher collectors (primary carer/secondary carer etc) works functionally now. All 3 columns on these pages should display properly now too. Middle column won't work functionally due to missing age DOB stuff. Tests for these pages will fail because I need to change them.

### DIFF
--- a/resources/views/store/partials/other_infoSP.blade.php
+++ b/resources/views/store/partials/other_infoSP.blade.php
@@ -8,7 +8,7 @@
     @if (!isset($family))
         <div class="user-control">
             <input type="checkbox" class="styled-checkbox @if($errors->has('consent'))invalid @endif" id="privacy-statement" name="consent" @if( old('consent') ) checked @endif/>
-            <label for="privacy-statement">Has the registration form been completed and signed?</label>
+            <label for="privacy-statement">Has the registration form been completed and signed?</label><br></br>
         </div>
         @if ( $errors->has('consent') )
         <div class="alert-message error" id="registration-alert">
@@ -45,7 +45,7 @@
         </a>
 
         @if (!isset($registration->family->leaving_on) )
-        <button class="remove long-button" type="button">Remove this family</button>
+        <button class="remove long-button" type="button">Remove this household</button>
         <div id="expandable" class="collapsed confirm-leaving">
             <div class="reason">
                 <label for="reason-for-leaving">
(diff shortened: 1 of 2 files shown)


resources/views/store/create_registration.blade.php
```diff
@@ -1,220 +1,34 @@
...
-        });
-        //remove invalid class when input is selected/tabbed to
-        $('#privacy-statement, #carer').on('click focus', function() {
-            $(this).removeClass("invalid");
-            // Remove relevent error message
-            var spanclass = $(this)[0].id + '-span';
-            $('#' + spanclass).addClass('collapsed');
-        });
-    </script>
+<script src="{{ asset('store/js/create_registration.js') }}"></script>
```
charlesstrange2, line 32: look at how mix is doing the copying and mangling of javascript into `public/`

resources/views/store/edit_registration.blade.php
```diff
@@ -4,304 +4,32 @@
...
-        });
-
-        $('#cancel').click(function (e) {
-            $('#expandable').addClass('collapsed');
-            $('#leaving').removeClass('expanded');
-            e.preventDefault();
-        });
-
-    </script>
+<script src="{{ asset('store/js/edit_registration.js') }}"></script>
```
charlesstrange2, line 33: similarly, look at how mix is mangling javascript into the public directories.

resources/views/store/partials/other_infoSP.blade.php
```diff
@@ -0,0 +1,70 @@
...
+        </div>
+        <button class="long-button submit" type="submit" formnovalidate>Save Changes</button>
+        <a href="{{ route("store.registration.voucher-manager", ['registration' => $registration ]) }}" class="link">
+            <div class="link-button link-button-large">
+                <i class="fa fa-ticket button-icon" aria-hidden="true"></i>Go to voucher manager
+            </div>
+        </a>
+
+        @if (!isset($registration->family->leaving_on) )
+        <button class="remove long-button" type="button">Remove this family</button>
```
charlesstrange2, line 48: check for "family"

charlesstrange2 (changes requested):
check for family names etc.

resources/views/store/partials/other_infoSP.blade.php
lauragurney: Done

resources/views/store/create_registration.blade.php
lauragurney: As discussed, we will be leaving this well alone for now!

PR #545, opened 2022-06-07
RegistrationPageTest fix

lauragurney:
This is the only test that can be fixed currently, as the remaining failures relate to being able to see age-col in EditPageTest which won't be functional until @charlesstrange2 has pushed the remaining age DOB changes


charlesstrange2 (approved):
okedoke ⛵

PR #550, opened 2022-06-10
Fix validation error.

neonleann:
https://trello.com/c/hn8heTqF/1866-invalid-date-warnings-not-disappearing-when-valid-date-is-entered

### DIFF
--- a/resources/views/store/partials/add_participant_form.blade.php
+++ b/resources/views/store/partials/add_participant_form.blade.php
@@ -30,7 +30,7 @@
     });
 
     // Clear error message
-    $('document').on('childInput:validated', function(e) {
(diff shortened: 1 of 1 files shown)


resources/views/store/partials/add_participant_form.blade.php
```diff
@@ -30,7 +30,7 @@
     });
 
     // Clear error message
-    $('document').on('childInput:validated', function(e) {
+    $(document).on('childInput:validated', function(e) {
```
charlesstrange2, line 33: aha! curse you, single quotes! 🧙

charlesstrange2 (approved):
✨ good spot

PR #554, opened 2022-06-10
1.14/1871 new registrationpagetest tests

neonleann:
https://trello.com/c/DrhVj0Mn/1871-new-registrationpagetest-tests

### DIFF
--- a/app/Http/Controllers/Service/Admin/SponsorsController.php
+++ b/app/Http/Controllers/Service/Admin/SponsorsController.php
@@ -245,7 +245,7 @@ class SponsorsController extends Controller
                 "name" => "DeductFromCarer",
                 "value" => -7,
                 "purpose" => "credits",
-                "entity" => "App\Child",
+                "entity" => "App\Family",
             ]),
             new Evaluation([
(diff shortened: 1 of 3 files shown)


lauragurney (approved):
Now that looks smashing. Great work. 🐻‍❄️

app/Http/Controllers/Service/Admin/SponsorsController.php
```diff
@@ -245,7 +245,7 @@ public static function socialPrescribingOverrides()
                 "name" => "DeductFromCarer",
                 "value" => -7,
                 "purpose" => "credits",
-                "entity" => "App\Child",
+                "entity" => "App\Family",
```
charlesstrange2, line 248: aha! good spot

charlesstrange2: this entire array feels like it should be a data item in `config/arc`
similarly the rules in the evaluatorfactory

we should make a card to extract them

charlesstrange2 (approved):
🏁 okedoke 🚗

PR #558, opened 2022-06-14
1.14/1880 allow secondary carers to be soft deleted

neonleann:
https://trello.com/c/XSEz2Lnq/1880-allow-secondary-carers-to-be-soft-deleted
I'm not sure that test is all that useful, but it works and it's there.


charlesstrange2 (approved):
:+1: lovely

PR #563, opened 2022-06-23
Fixing the eligibility issue.

neonleann:
https://trello.com/c/1jP2OyAk/1888-unable-to-change-the-healthy-start-or-nrpf-dropdowns
For some reason, the 'old' thing isn't working any more on those partials (I recall we had a similar problem with another variable). It's probably worth investigating at some point, but for the sake of releasing this week, this sorts the problem.


charlesstrange2 (approved):
Give it a go :-)

PR #571, opened 2022-08-02
Make No the NRPF default option.

neonleann:
https://trello.com/c/mLvE5T9m/784-nrpf-families-automatically-no-instead-of-yes-if-centres-arent-selecting-properly


charlesstrange2 (approved):
okedoke!

PR #572, opened 2022-08-02
HOTFIX - Fixes for P display and ID check box after error.

neonleann:
https://trello.com/c/Gcu7Z0mJ/783-hotfix-children-or-pregnancies-bug-in-the-store


lauragurney (approved):
Yes this works with L9

PR #574, opened 2022-08-05
HOTFIX: Prevent previous search from appearing

lauragurney:
https://trello.com/c/8QbRE61u/766-viewing-vouchers-created


neonleann (approved):
That will do the job.

PR #575, opened 2022-09-05
Fixing the Scottish tests.

neonleann:
https://trello.com/c/skh899qm/786-scottish-tests-failing

I think this will do it. It seems that we were just getting lucky for the first six months.


lauragurney (approved):
🏴󠁧󠁢󠁳󠁣󠁴󠁿 yes

PR #576, opened 2022-10-21
Minimal fix for requirements

charlesstrange2:
requires review prior to hot fixing
https://trello.com/c/pKHwCwRd/803-hotfix-h-f-and-v-on-prescription-data

Could have gone overboard with optimising these two, but i think it'll do for now.
The whole registration reports system could benefit from a template based approach.


neonleann (approved):
Excellent!

PR #600, opened 2023-03-07
1.15/1773 403 when changing centres vv e

lauragurney:
https://trello.com/c/XMgZbiz2/1773-403-when-changing-centres-vv-e


charlesstrange2 (approved):
okedoke :car:

PR #602, opened 2023-03-14
adding third party mandrill mailer to services

karalan:
All the other config changes occur in .env so will have to be done after new server is up during deploy.


lauragurney (approved):
✉️ 🐒

charlesstrange2 (approved):
🙈

PR #608, opened 2023-03-20
fix composer.lock and connect trader properly on collect.

charlesstrange2:
When a trader collects vouchers i had forgotten to associate the trader with that 😵‍💫 
Also fixed a `composer.lock` flub and added some comments


neonleann (approved):
🔐 Looks good.

PR #613, opened 2023-03-24
Tests for rejoin and view blade

lauragurney:
https://trello.com/c/OwZxG74t/1964-write-tests-for-new-rejoin-functionality

Some new tests, some edited tests and a correction to the SP reg download in CentreController that was missed previously. The test alerted me to that oversight 👏

### DIFF
--- a/tests/Feature/Store/RejoinPageTest.php
+++ b/tests/Feature/Store/RejoinPageTest.php
@@ -77,9 +77,9 @@ class RejoinPageTest extends StoreTestCase
         ;
         // Don't see the names of children in the page
         foreach ($this->children as $child) {
-            $this->dontsee('<td class="age-col">'. $child->getAgeString() .'</td>')
-                ->dontsee('<td class="dob-col">'. $child->getDobAsString() .'</td>')
-                ->dontseeElement('input[type="hidden"][value="'. $child->dob->format('Y-m') .'"]')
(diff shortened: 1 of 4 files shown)


tests/Feature/Store/RejoinPageTest.php
```diff
@@ -0,0 +1,128 @@
...
+    public function itWillAllowFamilyToRejoin()
+    {
+        $this->actingAs($this->centreUser, 'store')
+            ->visit(URL::route('store.registration.view', $this->registration->id))
+            ->press('Rejoin')
+            ->seePageIs(URL::route('store.registration.edit', [ 'registration' => $this->registration]));
+
+        $this->seeInDatabase('families', [
+            'id' => $this->family->id,
+            'rejoin_on' => Carbon::now()
```
neonleann, line 125: There's nothing wrong with this per se, but using plain old Carbon::now() can lead to intermittently failing tests when it decides the milliseconds don't match up. It's safest to use:
 `$knownDate = Carbon::create(2022, 5, 1, 12);` - pick your own favourite day and time
 `Carbon::setTestNow($knownDate);`
with `Carbon::setTestNow();` at the end of the test to reset things.

lauragurney: Done. Thanks for the tip, shall avoid in future tests!

PR #616, opened 2023-03-28
New seeder for a user with large amount of vouchers

lauragurney:
https://trello.com/c/LQieqBCK/1966-create-a-seeded-trader-with-a-large-history


neonleann (approved):
🌷

PR #619, opened 2023-03-29
Adding field to reg download

lauragurney:
https://trello.com/c/ZDSD9FsP/1438-reinstate-a-family-after-leaving-vv-eee

This is the work for calculating days spent on programme which is only used in the reg download.

### DIFF
--- a/app/Http/Controllers/Store/CentreController.php
+++ b/app/Http/Controllers/Store/CentreController.php
@@ -137,7 +137,8 @@ class CentreController extends Controller
             'dob' => 'd/m/Y',
             'join' => 'd/m/Y',
             'leave' => 'd/m/Y',
-            'eligible_from' => 'd/m/Y'
+            'eligible_from' => 'd/m/Y',
+            'rejoin' => 'd/m/Y',
         ], $dateFormats);
 
         // Get registrations decorated - may no longer be terribly efficient.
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Store/CentreController.php
```diff
@@ -234,6 +235,23 @@ private function getCentreFamilyRegistrationsSummary(array $centre_ids, $dateFor
             // Add our kids back in
             $row = array_merge($row, $kids);
 
+            // Calculate number of days on programme
+            if ($reg->family->leaving_on && !$reg->family->rejoin_on) {
+                $startDate = Carbon::parse($reg->created_at);
+                $leaveDate = Carbon::parse($reg->family->leaving_on);
+                $diff = $startDate->diffInDays($leaveDate);
+            } elseif ($reg->family->rejoin_on) {
```
neonleann, line 243: Is there any chance they might not have a leaving_on value? If there is, then you probably need to check it exists as well, so you don't try and do a parse on null.

app/Http/Controllers/Store/CentreController.php
```diff
@@ -417,6 +437,23 @@ private function getCentreSPRegistrationsSummary(array $centre_ids, $dateFormats
...
```
neonleann, line 445: Ditto what I said up there.

tests/Unit/Controllers/Store/CentreControllerTest.php
```diff
@@ -213,7 +214,8 @@ public function testItCanDownloadASocialPrescriptionsRegistrationsSpreadsheet()
             "Leaving Date",
             "Leaving Reason",
             "Rejoin Date",
-            "Leave Count"
+            "Leave Count",
+            "Days on programme"
         ];
         // The unexpected headers
         $unexpected_headers = [
```
neonleann, line 221: Are we not testing the actual functionality?

app/Http/Controllers/Store/CentreController.php
lauragurney: If they have a rejoin_on, they will have a leaving_on value because they have left once and then rejoined

neonleann (approved):
Family approved.

tests/Unit/Controllers/Store/CentreControllerTest.php
lauragurney: I've added to the existing card about fixing tests: https://trello.com/c/VCK9yl9L/1958-improve-fix-skipped-tests

PR #621, opened 2023-03-30
Participant not child for SP

lauragurney:
Again raised by Karl while testing https://trello.com/c/ZDSD9FsP/1438-reinstate-a-family-after-leaving-vv-eee


neonleann (approved):
🚸

PR #622, opened 2023-03-30
Changes required for SP too

lauragurney:
Picked up in testing. The SP blade was overlooked when these changes went into the family blade. This fixes it.

https://trello.com/c/3GwPEj7x/1991-rejoining-a-household-not-able-to-see-the-remove-household-button-after-rejoining

### DIFF
--- a/resources/views/store/partials/other_infoSP.blade.php
+++ b/resources/views/store/partials/other_infoSP.blade.php
@@ -43,27 +43,27 @@
                 <i class="fa fa-ticket button-icon" aria-hidden="true"></i>Go to voucher manager
             </div>
         </a>
-
-        @if (!isset($registration->family->leaving_on) )
-        <button class="remove long-button" type="button">Remove this household</button>
-        <div id="expandable" class="collapsed confirm-leaving">
-            <div class="reason">
-                <label for="reason-for-leaving">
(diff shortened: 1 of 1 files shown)


resources/views/store/partials/other_infoSP.blade.php
```diff
@@ -43,27 +43,27 @@
...
+            <button class="remove long-button" type="button">Remove this family</button>
+            <div id="expandable" class="collapsed confirm-leaving">
+                <div class="reason">
+                    <label for="reason-for-leaving">
+                        Reason for leaving
+                    </label>
+                    <select id="reason-for-leaving" name="leaving_reason" required>
+                        <option value="" disabled selected>Select a reason...</option>
+                        @foreach(Config::get('arc.leaving_reasons') as $reason)
+                            <option value="{{ $reason }}"> {{ $reason }}</option>
```
neonleann, line 57: Would you ever come here after a family has left and expect to see the reason? Because at the moment, it will only ever show 'Select a reason...'.

lauragurney: I don't think so? It's exactly the same as the other blade for non SP registrations.

neonleann (approved):
✔️

PR #623, opened 2023-03-31
1.15/1938 message when scanning vouchers vvv eee etags

charlesstrange2:
OK, this 

- adds Etag capability, so a client can be sent 304s if it asks for the same egat
- trusts the voucher current state a lot more
- fixes some tests that were failing because the functions were a bit off.

### DIFF
--- a/app/Http/Controllers/API/TraderController.php
+++ b/app/Http/Controllers/API/TraderController.php
@@ -185,10 +185,10 @@ class TraderController extends Controller
 
     /**
      * Helper to work out numbers of standard and SP vouchers.
-     * @param $vouchers
+     * @param ArrayAccess|array $vouchers
      * @return int[]
      */
-    public static function calculateProgrammeVoucherAmounts($vouchers): array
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/API/TraderController.php
```diff
@@ -203,60 +204,53 @@ public function calculateProgrammeVoucherAmounts($vouchers)
...
-                    $standard_areas[$voucher->sponsor->name] = 1;
-                } else {
-                    $standard_areas[$voucher->sponsor->name] += 1;
-                }
-            }
+        foreach ($vouchers as $voucher) {
+            $sponsorName = $voucher->sponsor->name;
+            $index = $voucher->sponsor->programme;
+            $programme_area_amounts[$index][$sponsorName] = isset($programme_area_amounts[$index][$sponsorName])
+                ? $programme_area_amounts[$index][$sponsorName] +1
```
neonleann, line 218: Not `+=1` ?

neonleann (approved):
Approved.

PR #626, opened 2023-04-03
1.15/1982 email selected payment record not firing and produces an error fix

charlesstrange2:
https://trello.com/c/K3LkELw6/1982-email-selected-payment-record-not-firing-and-produces-an-error

change the code to not trigger the front end queue handling code.


lauragurney (approved):
👍

PR #627, opened 2023-04-04
1992-403-error-whe-saving-changes-on-sp

lauragurney:
https://trello.com/c/Tn1F5pRA/1992-403-error-whe-saving-changes-on-sp

The HouseholdMember SP rule wasn't amended on the previous PR, this is the correction to ensure SP households see the correct voucher entitlement number

### DIFF
--- a/app/Services/VoucherEvaluator/Evaluations/HouseholdMember.php
+++ b/app/Services/VoucherEvaluator/Evaluations/HouseholdMember.php
@@ -22,7 +22,7 @@ class HouseholdMember extends BaseChildEvaluation
     {
         parent::test($candidate);
 
-        return ($candidate->family->leaving_on === null)
+        return ($candidate->leaving_on === null || $candidate->rejoin_on > $candidate->leaving_on)
             ? $this->success()
             : $this->fail()
         ;
--- a/composer.lock
+++ b/composer.lock
@@ -2731,26 +2731,27 @@
         },
         {
             "name": "league/oauth2-server",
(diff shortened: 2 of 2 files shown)


app/Services/VoucherEvaluator/Evaluations/HouseholdMember.php
```diff
@@ -22,7 +22,7 @@ public function test($candidate)
     {
         parent::test($candidate);
 
-        return ($candidate->family->leaving_on === null)
+        return ($candidate->leaving_on === null || $candidate->rejoin_on > $candidate->leaving_on)
```
neonleann, line 25: Two questions. 1) Will ->rejoin_on and ->leaving_on always exists on $candidate, such that it won't blow up because it can't find them, or end up making an incorrect comparison? 2) Given that ->rejoin_on and ->leaving_on are presumably dates, are they in a format that is comparable simply using '>'. Date comparisons can be tricky.

lauragurney: 1) Leaving_on will always exist and will be null on current registrations, which fulfils the first statement. If it's not null they will always have a rejoin_on because they rejoined. If neither, it doesn't matter, they are not active and don't need voucher entitlement calculated. 2) yes those dates can be compared with no issues

Previous PR for HouseholdExists rule is here https://github.com/neontribe/ARCVService/pull/625

neonleann (approved):
Okay.

PR #628, opened 2023-04-05
fix queue problems with user state

charlesstrange2:
https://trello.com/c/pl1UXkfG/1994-when-a-request-payment-job-fails-we-get-an-error-in-market

queues are hard :-(

we need to remind them we're logged in as a specific user.


lauragurney (approved):
Let's give it a whirl

PR #629, opened 2023-04-06
fix queue problems with timing and failing

charlesstrange2:
[2002-when-a-queue-job-fails-it-cant-record-t-the-failed-jobs-db](https://trello.com/c/GQvB0Qt1/2002-when-a-queue-job-fails-it-cant-record-t-the-failed-jobs-db)


neonleann (approved):
🌴

PR #630, opened 2023-04-11
1.15/1993 add etag support to market app

charlesstrange2:
https://trello.com/c/A13MknJ0/1993-add-etag-support-to-market-app

Added some more e-tag support service side
If you

... GET traders/{trader}/vouchers
... GET traders/{trader}/voucher-history
it will etag them

if you get them again with an if-none-match header with that etag AND there's no change in the response, it will just tell you 304
this stops multi-meg data transfers across the web.


neonleann (approved):
I see no problems here 🏷️.

PR #636, opened 2023-04-17
New Scottish and Southwark user seeds

lauragurney:
https://trello.com/c/2vjHBwgF/2006-create-seeded-users-for-scottish-and-southwark-rules

The seeders already created the relevant sponsors, centres, registrations etc, it was just missing the actual users needed for testing. This now means Karl won't have to manually create these users during each testing cycle.


neonleann (approved):
🌱

PR #637, opened 2023-04-17
stop using the xsrf cookie

charlesstrange2:
we don't use it.
https://trello.com/c/6sttDdCy/2013-remove-xsrftoken-cookie


neonleann (approved):
🍪

PR #646, opened 2023-05-16
fix: pass command as array other than string

tobybatch:


tests/Feature/Service/AdminResetTest.php
```diff
@@ -0,0 +1,74 @@
...
+                'redirect' => "0",
+                'personal_access_client' => 3,
+                'password_client' => 4,
+                'revoked' => 5,
+                'created_at' => "0",
+                'updated_at' => "0",
+            )
+        );
+        // Mock DB - DOES NOT WORK. I think the Process call spawns a new thread
+//        $mockDb = Mockery::mock(DatabaseManager::class, );
```
neonleann, line 48: I know this is commented out, but there's something missing after the comma here. I mention it just in case it had anything to do with it not working.

PR #647, opened 2023-05-19
Dev/docker

tobybatch:
No need for a full virtual box instance, just a docker instead. Has no effect on the homestead version


tobybatch:
> I don't know much about Docker but nothing here seems insane.

It means I don't need to run homestead and give it a bunch of memory. I'll explain containers sometime if you want.

PR #657, opened 2023-06-14
test: remove deps on external smtp servers

tobybatch:
I did this last night, not as part of the sprint but to make myself happy.

It just uses Mocks rather than actually sending the email. Better testing practice.


lauragurney (approved):
Fake it til you make it 👍

PR #741, opened 2024-02-15
Dev/add commitlint

tobybatch:
This will enforce https://commitlint.js.org/#/ style messages. It's just good practice and allows for better change log generation.


tobybatch:
Either run `yarn install` or `npx husky init` after pulling the merged version of this.

PR #755, opened 2024-03-04
Fix 977 fix tests causing workflow failures in GitHub

lauragurney:
https://trello.com/c/23anShkC/977-fix-tests-causing-workflow-failures-in-github

I switched the workflow back on in this branch on push and it all passed.

### DIFF
--- a/.github/workflows/laravel.yml
+++ b/.github/workflows/laravel.yml
@@ -2,9 +2,9 @@ name: Laravel
 
 on:
   workflow_dispatch:
-#  push:
-#   pull_request:
-#     branches: [ "1.16/release" ]
+  push:
+   pull_request:
+    branches: [ "1.16/release" ]
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/laravel.yml
```diff
@@ -2,9 +2,9 @@ name: Laravel
 
 on:
   workflow_dispatch:
-#  push:
-#   pull_request:
-#     branches: [ "1.16/release" ]
+  push:
+   pull_request:
+    branches: [ "1.16/release" ]
```
tobybatch, line 7: This should be just the `push` that will get all the conditions

```suggestion
  push:
  # pull_request:
  #  branches: [ "1.16/release" ]
```

tobybatch (changes requested):

tobybatch (approved):

PR #788, opened 2024-10-24
skunk: moves familes to anew centre

tobybatch:


app/Console/Commands/ArcSkunk.php
```diff
@@ -0,0 +1,57 @@
...
+
+        $c = Centre::where("id", $teviot_centre_id)->first();
+        $reg_associated_with_fam = $c
+            ->registrations()
+            ->withFullFamily()
+            ->get()
+            ->filter(function ($f) {
+                return in_array($f->family->centre_sequence, [4, 42, 44, 46, 39, 40, 41, 45]);
+            });
+        $bundles = $reg_associated_with_fam->first()->bundles;
```
charlesstrange2, line 43: And because if that, you want to iterate that array of registrations and fetch the reg->bundles for each.

Then iterate the bundles and set the disbursing centre id.

charlesstrange2 (changes requested):

app/Console/Commands/ArcSkunk.php
charlesstrange2: This version just explodes the first bundle of the first reg

PR #795, opened 2024-12-19
Upgrade prep reconcile into dev part2

charlesstrange2:
This is the process of uplifting _all the things_ to the most comfortable maximum we can for Laravel 9.

https://trello.com/c/Z14tNZ70/2136-fully-update-current-v9-laravel

@tobybatch you may find doing this one first is better
https://github.com/neontribe/ARCVService/pull/794


app/Voucher.php
```diff
@@ -136,11 +136,11 @@ public static function generateCodeRange($start, $end = "")
...
-                    $val,
-                    strlen($startMatch['number']),
-                    "0",
-                    STR_PAD_LEFT
-                );
+                        $val,
+                        strlen($startMatch['number']),
+                        "0",
+                        STR_PAD_LEFT
+                    );
```
tobybatch, line 143: Super snarky, is this indenting wrong?

tobybatch (approved):

app/Voucher.php
charlesstrange2: Snark is valid!

It look like phpstorm must have brought it to my attention /  fixed it by the time we hit the l10-11 PR.
image

PR #799, opened 2025-01-10
Upgrade l10 l11

charlesstrange2:
Body of changes for L10-L11

Mostly dealing with migrations and tests, as removal of DBAL and replacing it with native commands required rewrites.

Look at this first
https://github.com/neontribe/ARCVService/pull/797 

this GitGuardian doen't ike the `.env.ghactions`


config/database.php
```diff
@@ -10,56 +10,60 @@
...
+    | is supported by Laravel. You're free to add / remove connections.
     |
     */
 
     'connections' => [
 
         'sqlite' => [
             'driver' => 'sqlite',
-            'url' => env('DATABASE_URL'),
+            'url' => env('DB_URL'),
```
tobybatch, line 37: Isn't a bit risky? We'll need to change all the env files

charlesstrange2: When I went hunting I tlooka like the default is null when not specified. 

Which is fine, as currently it's local, not remote (in another container).

But... Normalising the config files will be important for when we do want to.

PR #802, opened 2025-01-17
fix: LargeVoucherSeeder history

charlesstrange2:
LargeVoucherSeeder needed a better history for the pending and reimbursed  vouchers.

This should fix that.

https://trello.com/c/o7kZhvWI/2152-largevouchersseeder-needs-a-full-voucher-history

### DIFF
--- a/app/Http/Controllers/API/TraderController.php
+++ b/app/Http/Controllers/API/TraderController.php
@@ -109,7 +109,7 @@ class TraderController extends Controller
                     // use inner join
                     ->join('voucher_states', 'vouchers.id', 'voucher_states.voucher_id')
                     ->where('voucher_states.to', 'payment_pending')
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/API/TraderController.php
```diff
@@ -109,7 +109,7 @@ public function showVoucherHistory(Trader $trader): JsonResponse
                     // use inner join
```
charlesstrange2, line 109: this is all IDE eagerly applying PHP CS fixer.

database/seeders/DatabaseSeeder.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
 namespace Database\Seeders;
```
charlesstrange2, line 3: Same for this one.

tobybatch (approved):

PR #814, opened 2025-02-07
2170 roll up repo

charlesstrange2:
https://trello.com/c/URCFxiwW/2170-roll-up-packages-e-vvv

Fixes:
- Laravel 11.40.0 - added tighter checking that classes that have classes that actually declare they have fields, should have those in the table. specifically this affects the soft deletes.

This became important beucase we had some data mangling as part of a pivot table migration that called CentreUsers, and at that point the CentreUser model delcared soft deletes, but gained those in a later migration.

I moved the soft delete addition into the CentreUser creation migration; this should not affect live migrations

- PHPStan removed some options in the config file. Larastan/Larastan should take care of most of the other stuff.


database/migrations/2018_07_24_111137_create_centre_users_table.php
```diff
@@ -24,6 +24,7 @@ public function up(): void
             $table->rememberToken();
             $table->integer('centre_id')->unsigned()->nullable();// Not all Users may have a CC e.g. Admins
             $table->timestamps();
+            $table->softDeletes();
```
tobybatch, line 27: Should you be updating an existing migration?

Shouldn't this be a new migration.

tobybatch (approved):

database/migrations/2018_07_24_111137_create_centre_users_table.php
charlesstrange2: See above;

- this doesn't affect live, live has already run the migrations.
- this affect staging and local, which blow up because the model says soft deletes, but a later migration manipulates data before the softdeletes are added.
- can't add a new migration, as it will be run after the problem occurs.

Solution is to fold the later soft delete migration into the original creation, respecting intent but before the data manipulation happens.

PR #860, opened 2026-04-02
1.20/2263 admin portal download area ee vvv

charlesstrange2:
The Admin service needs a way to pull a CSV file of the secret emails and telnos
- best to stream this
- it has some tests to sanity check the answers.

** when analysing test outputs, I see there's a transient bug in the crypted storage system: it looks like sometimes we can overwrite the plantext with ciphertext when saving, leading to unrecoverable text :thinking:  i shall fix this next week **

On the way
- made the sidebar nav collapse the list of functions by default; should make for a better experience
- rewrote the routes files (data, store, service, api, web) to the newer l12 format to avoid deprecation messages in logs. 
- protected the data routes better with a debug gate as well as admin auth requirments 
- especially the reset data route that was very broken by l12 updates (and also, a bit dangerous
  - this is now a _job_ that is called
  - overwriting the old passport key in the `.env` was happening after every test, this is now isolated as a mockable service.
  
https://trello.com/c/5W0Pk4X9/2263-admin-portal-download-area-ee-vvv
https://trello.com/c/wCbrqQm3/2265-bug-the-data-route-is-broken
https://trello.com/c/T8obHbVF/2264-bug-update-routes

### DIFF
--- a/app/Http/Controllers/Service/Data/FamilyContactsController.php
+++ b/app/Http/Controllers/Service/Data/FamilyContactsController.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace App\Http\Controllers\Service\Data;
+
+use App\Carer;
+use App\Http\Controllers\Controller;
+use Illuminate\Http\Request;
+use Illuminate\Support\Carbon;
+use Symfony\Component\HttpFoundation\StreamedResponse;
+
+class FamilyContactsController extends Controller
+{
+    public function __invoke(Request $request): StreamedResponse
+    {
+        $now = Carbon::now()->format('YmdHis');
+        return response()->streamDownload(function () {
+
+            $output = fopen('php://output', 'wb');
+
+            fputcsv($output, ['Rvid', 'Name', 'Email', 'Telno', 'Centre', 'Area']);
+
+            Carer::query()
+                ->whereNotNull('emailsecret')
+                ->whereNotNull('telnosecret')
+                ->with(['family.initialCentre.sponsor'])
+                ->lazyById(200)
+                ->chunk(200)
+                ->each(function ($chunk) use ($output) {
+                    $chunk->each(function (Carer $carer) use ($output) {
+                        fputcsv($output, [
+                            $carer->family?->Rvid,
+                            $carer->name,
+                            $carer->emailsecret->reveal(),
+                            $carer->telnosecret->reveal(),
+                            $carer->family?->initialCentre?->name,
+                            $carer->family?->initialCentre?->sponsor?->name,
+                        ]);
+                    });
+                    ob_flush();
+                    flush();
+                });
+
+            fclose($output);
+        }, "carers-export_$now.csv", [
+            'Content-Type' => 'text/csv',
+            'Cache-Control' => 'no-cache',
+            'X-Content-Type-Options' => 'nosniff',
+        ]);
+    }
+}
--- a/app/Http/Controllers/Service/Data/FamilyContactsController.php
+++ b/app/Http/Controllers/Service/Data/FamilyContactsController.php
@@ -21,7 +21,7 @@ class FamilyContactsController extends Controller
 
             Carer::query()
                 ->whereNotNull('emailsecret')
-                ->whereNotNull('telnosecret')
+                ->orWhereNotNull('telnosecret')
                 ->with(['family.initialCentre.sponsor'])
                 ->lazyById(200)
                 ->chunk(200)
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -410,7 +410,7 @@ button.remove {
             padding: 1em 2em;
             &:hover,
             &:active {
-                background-color: #f5f5f5;
+                background-color: $arc_gray;
             }
         }
 
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -398,7 +398,7 @@ button.remove {
     flex: 0 0 300px;
     font-weight: 600;
     background-color: $arc_white;
-    border-right: 1px solid #ededed;
+    border-right: 1px solid $arc_border;
 
     ul {
         padding-left: 0;
--- a/resources/assets/sass/app.scss
+++ b/resources/assets/sass/app.scss
@@ -343,29 +343,6 @@ span.user-array {
   min-height: 100%;
 }
 
-#sidebar {
-  flex: 0 0 300px;
-  font-weight: 600;
-  background-color: $arc_white;
-  border-right: 1px solid #ededed;
-  ul {
-    padding-left: 0;
-    margin-top: 1em;
-    li {
-      color: #656565;
-      display: block;
-      padding: 1em 2em;
-      &:hover,
-      &:active {
-        background-color: #f5f5f5;
-      }
-    }
-    span {
-      padding-right: 10px;
-    }
-  }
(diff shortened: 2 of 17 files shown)


app/Http/Controllers/Service/Data/FamilyContactsController.php
```diff
@@ -0,0 +1,50 @@
...
+        $now = Carbon::now()->format('YmdHis');
+        return response()->streamDownload(function () {
+
+            $output = fopen('php://output', 'wb');
+
+            fputcsv($output, ['Rvid', 'Name', 'Email', 'Telno', 'Centre', 'Area']);
+
+            Carer::query()
+                ->whereNotNull('emailsecret')
+                ->whereNotNull('telnosecret')
```
colinstewart-dxw: From the acceptance criteria on Trello:
> As an admin user the file contains details for users who have **at least 1 of email or telno**

(**my emphasis**)

Confidence check me: Should `orWhereNotNull()` be used here so it's OR rather than AND?
```php
->whereNotNull('emailsecret')
->orWhereNotNull('telnosecret')
```

resources/assets/sass/app.scss
```diff
@@ -416,3 +393,94 @@ tr.disabled {
 button.remove {
   background-color: #d2232a;
 }
+
+#sidebar {
+    flex: 0 0 300px;
+    font-weight: 600;
+    background-color: $arc_white;
+    border-right: 1px solid #ededed;
```
colinstewart-dxw:
```suggestion
    border-right: 1px solid $arc_border;
```

resources/assets/sass/app.scss
```diff
@@ -416,3 +393,94 @@ tr.disabled {
...
+
+    ul {
+        padding-left: 0;
+        margin-top: 1em;
+
+        li {
+            color: #656565;
```
colinstewart-dxw, line 408: Doesn't look like there's a SASS variable for this colour. The "nearest" is `$text-color: #636b6f`, which is _kind of close_. Should that variable be used here, another variable be added for `#656565`, or does it not matter?

resources/assets/sass/app.scss
```diff
@@ -416,3 +393,94 @@ tr.disabled {
...
+            display: block;
+            padding: 1em 2em;
+            &:hover,
+            &:active {
+                background-color: #f5f5f5;
```
colinstewart-dxw:
```suggestion
                background-color: $arc_gray;
```

resources/assets/sass/app.scss
colinstewart-dxw: Same question for `#999` and such later in the changes.

resources/views/service/includes/sidebar.blade.php
```diff
@@ -1,35 +1,210 @@
...
+                </li>
+                <li class="sidebar-item sidebar-item--danger">
+                    <a href="{{ route('data.reset') }}">
+                        <span class="glyphicon glyphicon-warning-sign"></span>
+                        Reset data
+                    </a>
+                </li>
+            </div>
+        </ul>
+    @endcan
```
colinstewart-dxw, line 208: Possible opportunity for creating sidebar, sidebar heading and sidebar item components here, albeit maybe not possible within the current time/scope/budget.

routes/api.php
```diff
@@ -1,102 +1,69 @@
...
-     * new voucher transition routes
-     */
-    Route::post('vouchers/transitions', [
-        'as' => 'api.vouchers.transition-responses.store',
-        'uses' => 'TransitionController@store',
-    ])->middleware('can:collect,App\Voucher');
+    // Legacy transition for old clients
+    Route::post('vouchers', [VoucherController::class, 'legacyTransition'])
+        ->name('api.voucher.transition')
+        ->middleware('can:collect,App\Voucher');
```
colinstewart-dxw, line 57: No opinion, just a thought: Is it worth concatenating `Voucher:class` instead of hardcoding `App\Voucher` here and in `app.vouchers.transition-responses.store`?

routes/store.php
```diff
@@ -1,204 +1,147 @@
...
+/*
+|--------------------------------------------------------------------------
+| Store Routes
+|--------------------------------------------------------------------------
+*/
+
+// Authentication
+Route::get('login', [LoginController::class, 'showLoginForm'])->name('store.login');
+Route::post('login', [LoginController::class, 'login']);
+Route::post('logout', [LoginController::class, 'logout'])->name('store.logout');
```
colinstewart-dxw: Does this need auth middleware to align with other `logout` routes?

tests/Services/EnvWriterTest.php
```diff
@@ -0,0 +1,160 @@
...
+    {
+        $original = "APP_NAME=MyApp\nAPP_ENV=local\n";
+        $this->writeEnv($original);
+
+        $this->makeWriter()->updateKey('PASSWORD_CLIENT_SECRET', 'new-secret');
+
+        $this->assertSame($original, $this->readEnv());
+    }
+
+    public function testUpdatesKeyWithEmptyCurrentValue(): void
```
colinstewart-dxw, line 96: Checking: Does your convention prefer individual test methods rather than using data providers for similar tests?

app/Http/Controllers/Service/Data/FamilyContactsController.php
charlesstrange2: yep, i flubbed implementing that properly. fixed

tests/Services/EnvWriterTest.php
charlesstrange2: We learned to love data providers for doing things like fuzzing the validation rules to save time.
Claude scaffold-ed a lot of these tests, which also saves time.

resources/assets/sass/app.scss
charlesstrange2: Many years ago there was a style guide, including colours that matched the ARC branding nicely.
I'm not a CSS/design brain, so i'm inclined to leave it (mostly) alone until they choose to do a grand re-branding.

resources/views/service/includes/sidebar.blade.php
charlesstrange2: It crossed my mind, while reorganising it all but 
- this is the only place we use it 
- it's relatively clear as-is 

and it was a bit of extra juice to make the sidebar manageable (as it was creeping down the page a bit), so i'l already spending budget i probably shouldn't :sweat_smile:

routes/api.php
charlesstrange2: A task for whenever (l13 upgrade next year?) we get round to shifting the models into an app/models directory and namespace per the newer directory recommendations, which will make a whole pile of ancillary changes everywhere.

resources/assets/sass/app.scss
colinstewart-dxw: Makes sense, thanks for the context!

resources/views/service/includes/sidebar.blade.php
colinstewart-dxw: Haha, no worries!

tests/Services/EnvWriterTest.php
colinstewart-dxw: Makes sense, thanks!

colinstewart-dxw (approved):
Thanks for the context and updates, this looks good to me!

PR #863, opened 2026-04-05
1.20/2189-bug-permenently-deleting-a-centre-user-ee-vv_part2

charlesstrange2:
https://trello.com/c/ytZJt1yn/2189-bug-permenently-deleting-a-centre-user-ee-vv

- converted the delete function in the controller to a retire function
- since we're not deleting any more we keep the relations
- updated the edit page to be "reitre" not "delete"
- updated the route name to be "retire" not "delete"
- updated a whole bunch of tests

### DIFF
--- a/resources/views/service/centreusers/edit.blade.php
+++ b/resources/views/service/centreusers/edit.blade.php
@@ -80,10 +80,10 @@
             <button type="submit" id="toggleWorker">@empty($worker->deleted_at)Disable worker @else Enable worker @endif</button>
         </form>
         @if($worker->deleted_at)
-        <form role="form" id="deleteForm" class="styled-form" method="GET"
-            action="{{ route('admin.centreusers.delete', ['id' => $worker->id]) }}">
+        <form role="form" id="retireForm" class="styled-form" method="GET"
+            action="{{ route('admin.centreusers.retire', ['id' => $worker->id]) }}">
             {!! csrf_field() !!}
-          <button type="submit" id="deleteWorker" class="remove">Delete worker</button>
+          <button type="submit" id="retireWorker" class="remove">Retire worker</button>
         </form>
         @endif
         <script>
@@ -175,12 +175,12 @@
                     // show the boxes when we load
                     buildCheckboxes();
                 });
-                $('#deleteWorker').on('click', function (evt) {
+                $('#retireWorker').on('click', function (evt) {
                   evt.preventDefault();
-                  var deleteForm = document.getElementById('deleteForm');
-                  var areYouSure = confirm('Are you sure you want to PERMANENTLY delete this worker account?');
+                  var retireForm = document.getElementById('retireForm');
+                  var areYouSure = confirm('Are you sure you want to PERMANENTLY retire this worker account?');
                   if (areYouSure === true) {
-                    deleteForm.submit();
+                    retireForm.submit();
                   };
                 });
         </script>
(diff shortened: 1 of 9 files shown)


resources/views/service/centreusers/edit.blade.php
```diff
@@ -80,10 +80,10 @@ class="{{ $errors->has('worker_centre') ? 'error' : '' }}" required>
             <button type="submit" id="toggleWorker">@empty($worker->deleted_at)Disable worker @else Enable worker @endif</button>
         </form>
         @if($worker->deleted_at)
```
colinstewart-dxw, line 82: If I'm reading this right:
- "Disable" = Softly delete
- "Enable" = Restore from soft deletion
- "Delete" = Permanently delete, only available for already disabled Centre users - This is being replaced by "Retire"
- "Retire" = Softly delete, anonymise and prevent restoration

In `Retirable::retire()`, `$this->delete()` is called "if we're not already soft deleted, do that too". This suggests a change in behaviour - that there is no longer a requirement to disable a user before retiring them.

If that behavioural change is intended, then it looks like this `@if($worker->deleted_at)` condition should be removed. If it's not intended, is there a known situation in which the user is being retired without already being soft deleted?

colinstewart-dxw: That said, `CentreUsersController::retire()` still only runs on already trashed (disabled) users (`$centreUser = CentreUser::onlyTrashed()->findOrFail($id);`), and the Acceptance Criteria in [the Trello card](https://trello.com/c/ytZJt1yn/2189-bug-permenently-deleting-a-centre-user-ee-vv) states:
> As an Admin, I can "retire" a disabled account, so that i can be sure the user's PD is no loger avalable.

So it sounds like the question should be:
- Should the "safety" of `$this->delete()` in `Retirable::retire()` actually be there, or should an error be thrown if someone has somehow triggered a retire on a user who hasn't been disabled first?

tests/Unit/Controllers/Service/Admin/CentreUserControllerTest.php
```diff
@@ -131,26 +128,24 @@ public function testItCanDisableACentreUser(): void
...
-            ->see('This worker is <i>disabled</i>')
-            ->seeInElement('#toggleWorker', ' Enable worker')
-            // There is a delete button
-            ->seeElement('#deleteWorker')
-        ;
+            ]);
+
+        // Record is soft-deleted but still present.
+        $this->assertNull(CentreUser::find($cu->id));
+        $this->assertNotNull(CentreUser::withTrashed()->find($cu->id)->deleted_at);
```
colinstewart-dxw: Confidence check me: If there's a bug and the user isn't found within `withTrashed()`, `find($cu->id)` will return `null`, and there'll be an error for trying to access the `deleted_at` property on `null`. It's therefore safer to check that `CentreUser::withTrashed()->find($cu->id)` is not `null`, even if there's also a follow-up assertion to check that `deleted_at` is not `null`.

Sidenote: This might just be a coding style matter and if so then no worries. In case it's not a coding style matter: passing a last argument (`$message`) can add context to a failing assertion:
```php
$this->assertNull(CentreUser::find($cu->id), 'The user was not soft-deleted');
$this->assertNotNull(CentreUser::withTrashed()->find($cu->id)->deleted_at, 'The user cannot be found in the trashed collection');
```

resources/views/service/centreusers/edit.blade.php
charlesstrange2: Yep, your mental model is correct and your suggestion works better, will amend.

tests/Unit/Controllers/Service/Admin/CentreUserControllerTest.php
charlesstrange2: I'll remove the deleted_at check - it's the focus of withTrashed, so checking is kinda redundant.

I've not considered messages before, as generally the IDE/debugger can tell me why it's borked. :thinking: 
Toby and i use the same setup but i'll add it as a tech debt consideration.

colinstewart-dxw (approved):

PR #864, opened 2026-04-08
2254 amend centres so they can be set to collect vouchers eee vvv

charlesstrange2:
part of:
https://trello.com/c/wXGqdOd5/2254-amend-centres-so-they-can-be-set-to-collect-vouchers-eee-vvv

- migration to add flag that indicates if a Centre can "collect" it's own vouchers or not
- amended to model to be aware of it
- made a factory for convenience 
- updated tests to check it

### DIFF
--- a/app/Centre.php
+++ b/app/Centre.php
@@ -27,7 +27,7 @@ class Centre extends Model
      * @var array
      */
     protected $fillable = [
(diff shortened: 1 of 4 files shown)


colinstewart-dxw (approved):
LGTM

PR #866, opened 2026-04-09
1.20/2254 amend centres so they can be set to collect vouchers eee vvv part 2 UI

charlesstrange2:
https://trello.com/c/wXGqdOd5/2254-amend-centres-so-they-can-be-set-to-collect-vouchers-eee-vvv

This adds a checkbox to the create and edit pages that will set the flag on the centre model

on the way:
- modernised the centre creation page
- used that as a base to update the utterly shonky centre edit page
- updated the validation rules for both
- simplified the store and update methods
- fixed some routes
- nudged some css to do with vertically positioning label text against the checkboxes
- added some tests to cover the new cases
- fixed some tests to do with prefixes that passing for the wrong reasons (the "EXISTS" prefix was too long, but that passed _because_ it was triggering a string length rule, not for whatever reason the test needed)

### DIFF
--- a/resources/views/service/centres/edit.blade.php
+++ b/resources/views/service/centres/edit.blade.php
@@ -80,7 +80,7 @@
                     <label for="can_collect">Can Collect Vouchers</label>
                     @include('service.partials.validationMessages', ['inputName' => 'can_collect'])
                 </div>
-                <button type="submit" id="createCentre">Save</button>
+                <button type="submit" id="updateCentre">Save</button>
             </form>
         </div>
     </div>
--- a/resources/views/service/centres/edit.blade.php
+++ b/resources/views/service/centres/edit.blade.php
@@ -1,42 +1,88 @@
 @extends('service.layouts.app')
 @section('content')
 
-<div id="container">
-    @include('service.includes.sidebar')
-    <div id="main-content">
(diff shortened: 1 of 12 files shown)


resources/views/service/centres/edit.blade.php
```diff
@@ -1,42 +1,88 @@
...
+                    <input type="checkbox"
+                           id="can_collect"
+                           name="can_collect"
+                           class="styled-checkbox @error('can_collect') error @enderror"
+                        @checked(old('can_collect', $centre->can_collect))
+                    >
+                    <label for="can_collect">Can Collect Vouchers</label>
+                    @include('service.partials.validationMessages', ['inputName' => 'can_collect'])
+                </div>
+                <button type="submit" id="createCentre">Save</button>
```
colinstewart-dxw: Checking: The `id` was previously `save`, however it's `createCentre` here despite being the edit template. Is that an intentional change or a copy/paste gotcha?

colinstewart-dxw (approved):
One clarifying question but otherwise looks good to me!

resources/views/service/centres/edit.blade.php
charlesstrange2: yep, copy / paste. fixed

PR #867, opened 2026-04-10
1.20/2268 centres can have integrated markets

charlesstrange2:
https://trello.com/c/o6DBaEQJ/2268-centres-can-have-integrated-markets-ee-vvv
- setup a schema to make sure markets can belong to a centre if they need to
- update the models
- add an observer (new stunt: annotations on classes to register, rather than in app service provider)
- add a service to fudge creating a minimally viable market and trader if the centre needs one
- write some tests for the models, observer, service.

### DIFF
--- a/app/Services/CentreCollectionMarketService.php
+++ b/app/Services/CentreCollectionMarketService.php
@@ -8,34 +8,29 @@ use App\Trader;
 
 class CentreCollectionMarketService
 {
-    public function ensureMarket(Centre $centre): Market
+    public function ensureTradingMarket(Centre $centre): void
     {
-        // withTrashed guards against the edge case where someone
-        // manually soft-deleted the market outside normal flows.
-        $market = Market::withTrashed()
-            ->where('centre_id', $centre->id)
-            ->first();
-
-        if ($market) {
-            // Restore if soft-deleted
-            if ($market->trashed()) {
-                $market->restore();
-            }
-        } else {
-            $market = Market::create([
-                'name' => $centre->name . ' Collection',
-                'location' => $centre->name,
-                'centre_id' => $centre->id,
-                'sponsor_id' => $centre->sponsor->id,
-                'payment_message' => '',
-            ]);
-
-            Trader::create([
-                'name' => $centre->name . ' Collection Trader',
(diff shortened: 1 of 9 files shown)


tests/Unit/Services/CentreCollectionMarketServiceTest.php
```diff
@@ -0,0 +1,100 @@
...
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->service = new CentreCollectionMarketService();
+        $this->centre = factory(Centre::class)->create(['can_collect' => false]);
+    }
+
+    public function testItCreatesAMarketWhenNoneExists(): void
+    {
+        $this->service->ensureTradingMarket($this->centre);
```
colinstewart-dxw, line 28: Might be worth confirming none exists at the start before checking that a new one is created, just to make sure the initial test environment is definitely as expected.

Same for the trader test below and similar for checking X markets/traders exist in other tests here.

app/Services/CentreCollectionMarketService.php
```diff
@@ -0,0 +1,36 @@
...
+            'name' => $centre->name . ' (Internal)',
+            'location' => $centre->name,
+            'centre_id' => $centre->id,
+            // assign it the centre's sponsor/area
+            'sponsor_id' => $centre->sponsor->id,
+            // don't need one of these, the payment recommendation will be automated
+            'payment_message' => '',
+        ]);
+
+        // ... and it's trader
```
colinstewart-dxw:
```suggestion
        // ... and its trader
```

Yes I hate myself for this one too.

colinstewart-dxw (approved):
Just a couple of comments, nothing major. Looks good to me otherwise!

tests/Unit/Services/CentreCollectionMarketServiceTest.php
charlesstrange2: reckon with RefreshDatabase we can be relatively confident it's empty on test start.

colinstewart-dxw: True

PR #870, opened 2026-04-13
refactor: modernise bundlecontroller and bundle model

charlesstrange2:
https://trello.com/c/8eNGK9tg/2252-create-a-new-collection-confirmation-control-eee-vvv
this (part 1) is just refactoring, and fixing the bundle and bundlecontroller classes

### DIFF
--- a/app/Http/Controllers/Store/BundleController.php
+++ b/app/Http/Controllers/Store/BundleController.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 namespace App\Http\Controllers\Store;
 
 use App\Bundle;
--- a/app/Http/Controllers/Store/BundleController.php
+++ b/app/Http/Controllers/Store/BundleController.php
@@ -1,367 +1,300 @@
 <?php
 
+declare(strict_types=1);
+
 namespace App\Http\Controllers\Store;
 
 use App\Bundle;
 use App\Carer;
 use App\Centre;
 use App\Family;
-use App\Voucher;
-use App\Registration;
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Store/BundleController.php
```diff
@@ -1,367 +1,300 @@
...
-                function ($value) {
-                    return !empty($value);
-                }
+        // --- Sync voucher codes if supplied ---
+        if ($request->exists('vouchers')) {
+            $rawCodes = array_filter(
+                $request->input('vouchers', []),
+                static function (mixed $v): bool {
+                    return !empty($v);
+                },
```
colinstewart-dxw: This callback appears to be equivalent to the default PHP behaviour for `array_filter()` when no callback is passed

app/Http/Controllers/Store/BundleController.php
```diff
@@ -1,367 +1,300 @@
...
+                $inaccessible[] = $voucher->code;
             }
-            // Otherwise, sure, return to the new view.
-            return redirect($successRoute)
-                ->with('message', $message);
         }
+
+        $messages = [];
+
+        if ($relevant !== []) {
```
colinstewart-dxw: Just flagging a small inconsistency: Elsewhere in the codebase, `empty($someArray)` is used, whereas this tests against an empty array. It's just a small issue of convention which may or may not be of note to you. Thought it was worth mentioning anyway.

colinstewart-dxw (approved):
Haven't spotted blockers, just left a couple of thoughts.

PR #872, opened 2026-04-16
1.20/2252 create a new collection confirmation control eee vvv partials

charlesstrange2:
https://trello.com/c/8eNGK9tg/2252-create-a-new-collection-confirmation-control-eee-vvv

Largely this chops up the voucher manager page so it's more manageable, and uses more modern blade constructs.
the rest is moving related pages into folders so that the store views directory contents are organised better, reducing the large number of blades in one place

On the way
- I missed an email that needs a proper address for smoke testing password resets etc.


resources/views/store/edit_registration.blade.php
```diff
@@ -4,33 +4,35 @@
...
-    <div class="content flex">
-        <form action="{{ URL::route("store.registration.update",['registration' => $registration]) }}" method="post" class="full-height">
-            {{ method_field('PUT') }}
-            {!! csrf_field() !!}
+    @if ($programme !== 0)
+        <div class="content flex">
+            <form action="{{ URL::route("store.registration.update",['registration' => $registration]) }}" method="post"
+                  class="full-height">
+                {{ method_field('PUT') }}
+                {!! csrf_field() !!}
```
colinstewart-dxw: Is it worth updating these to `@method('PUT')` and `@csrf` respectively while we're here? Same for the form below.

Also the `URL::route()` to `route()` and such.

colinstewart-dxw (approved):
One question, no blockers. LGTM 👍

resources/views/store/edit_registration.blade.php
charlesstrange2: indeed! updated

PR #873, opened 2026-04-16
1.20/2269 bug dont let people use all centres

charlesstrange2:
https://trello.com/c/u7u7vMMX/2269-bug-dont-let-people-use-all-centres

rolls back a hack to allow people to search all their centres by fudging the "which centre are we logged into" select box.
turns out this _kinda_ worked but had some odd side effects later on when recording where vouchers were handed out in the few cases where families turn up at the wrong centre.

This involved rewriting the search page's index function so it would work better, and the masthead/session state control

... and then a whole bunch of new tests. 

[I did have a big session with various LLMs to try to sort out the indexing functions and tests.]


app/Http/Controllers/Store/RegistrationController.php
```diff
@@ -13,185 +13,156 @@
...
+        $neighbourCentreIds = $user->relevantCentres()->pluck('id');
+
+        $baseQuery = Registration::query()
+            ->withPrimaryCarer()
+            ->whereIn('registrations.centre_id', $neighbourCentreIds);
+
+        if (
+            $user->centres->count() > 1 &&
+            $request->boolean('filter_by_centre') &&
+            ($centreId = session('CentreUserCurrentCentreId'))
```
colinstewart-dxw: Query for future reviewing: Should I flag assignment in conditionals or is it a common enough pattern in the project/other projects at Neontribe?

app/Http/Controllers/Store/SessionController.php
```diff
@@ -12,7 +12,7 @@ public function update(StoreUpdateSessionRequest $request): RedirectResponse
     {
         // Set session
         session(['CentreUserCurrentCentreId' => $request->input('centre')]);
-        // redirect to a specific place
-        return redirect()->route('store.registration.index');
+        // redirect to the prior route
+        return redirect(url()->previous());
```
colinstewart-dxw: Can also use `redirect()->back()` or `back()` unless you have a preference towards `redirect(url()->previous())`

app/Registration.php
```diff
@@ -135,7 +132,7 @@ public function centre(): BelongsTo
      *
      * @return Model
```
colinstewart-dxw:
```suggestion
```

app/Registration.php
```diff
@@ -135,7 +132,7 @@ public function centre(): BelongsTo
...
      */
-    public function currentBundle()
+    public function currentBundle(): Model
```
colinstewart-dxw: Seems to only return `Bundle`, or would you prefer to keep `Model` for backward compatibility or another reason?

app/Registration.php
```diff
@@ -200,4 +193,43 @@ public function lastBundle(): HasOne
             ->whereNotNull('disbursed_at')
             ->latest('disbursed_at');
     }
+
+    /**
+     * Join the primary carer (MIN id per family) directly into the query,
+     *  so we can filter and sort by carer name in SQL rather than PHP.
```
colinstewart-dxw:
```suggestion
     * Join the primary carer (MIN id per family) directly into the query,
     * so we can filter and sort by carer name in SQL rather than PHP.
```

app/Registration.php
```diff
@@ -200,4 +193,43 @@ public function lastBundle(): HasOne
...
+
+    public function scopeOrderByCarerName(Builder $query, bool $descending = false): Builder
+    {
+        return $query->orderBy('carers.name', $descending ? 'desc' : 'asc');
+    }
+
+    public function scopeFilterByCarerName(Builder $query, string $term): Builder
+    {
+        return $query
+            ->where('carers.name', 'LIKE', "%{$term}%")
```
colinstewart-dxw: `whereLike()` [Ref](https://laravel.com/docs/12.x/queries#additional-where-clauses) could be used here

resources/views/store/partials/masthead.blade.php
```diff
@@ -1,52 +1,55 @@
...
-    @if (!Auth::guest())
+
+    @auth
         <div class="header-section">
             <ul>
-                <li>User: {{ Auth::user()->name }} </li>
-                @if ( config('app.env') !== 'production' )
+                <li>User: {{ Auth::user()->name }}</li>
+
+                @if (app()->environment() !== 'production')
```
colinstewart-dxw: Take/Leave: This is fine but offering a couple of other options:

```blade
@if (!app()->environment('production'))

@endif
```

```blade
@unless (app()->environment('production'))

@endunless
```

resources/views/store/index_registration.blade.php
```diff
@@ -4,93 +4,127 @@
...
 
-    @include('store.partials.navbar', ['headerTitle' => 'Search for a ' . ($programme ? 'household' : 'family')])
-    @includeWhen(Session::has('message'), 'store.partials.success')
+    @include('store.partials.navbar', [
+        'headerTitle' => 'Search for a ' . ($programme ? 'household' : 'family')
+    ])
+
+    @if (session('message'))
+        @include('store.partials.success')
+    @endif
```
colinstewart-dxw: Looks like you're moving away from `@includeWhen` here. I think there were some changes in #872 that retained `@includeWhen`. If you're moving away from `@includeWhen`, I'm assuming that other instances will be picked up in a later pass?

resources/views/store/index_registration.blade.php
```diff
@@ -4,93 +4,127 @@
...
-                                        class="fuzzy-text-on"
-                                   @endif
-                                   onClick="setExactSearch()">Exact</a>
-                                <a href="#"
-                                   id="fuzzy-search-fuzzy"
-                                   @if ($fuzzy)
-                                        class="fuzzy-text-on"
-                                   @endif
-                                   onClick="setFuzzySearch()">Fuzzy</a>
+                                <a href="#" id="fuzzy-search-exact"
```
colinstewart-dxw, line 74: Likely OOS for this PR but just checking: If these links always have `#` and are JS-controlled, should they actually be `<button>` elements with `appearance: none` and some base text styling to align with `<a>` elements?

resources/views/store/index_registration.blade.php
```diff
@@ -113,30 +145,36 @@ class="link inline-link-button">
...
+
         <div>
             {{ $registrations->links() }}
-            Showing {{($registrations->currentPage()-1)* $registrations->perPage()+($registrations->total() ? 1:0)}} to {{($registrations->currentPage()-1)*$registrations->perPage()+count($registrations)}}  of  {{$registrations->total()}}  Results
+            Showing
+            {{ ($registrations->currentPage() - 1) * $registrations->perPage() + ($registrations->total() ? 1 : 0) }}
+            to
+            {{ ($registrations->currentPage() - 1) * $registrations->perPage() + count($registrations) }}
+            of
+            {{ $registrations->total() }}
```
colinstewart-dxw, line 153: Does this produce the same or different output as you have here?

```blade
Showing
{{ $registrations->firstItem() }}
to
{{ $registrations->lastItem() }}
of
{{ $registrations->total() }}
```

resources/views/store/partials/masthead.blade.php
charlesstrange2: We also have a gate made this sprint, which encapsulates  "a dev might need this while debugging" @can('take-developer-actions")

app/Http/Controllers/Store/RegistrationController.php
charlesstrange2: there's no sepcific policy on this pattern. If it's unclear, or obtusely terse, flag it.

app/Registration.php
charlesstrange2: I noticed this too - fixed in the next PR :sweat_smile: 

https://github.com/neontribe/ARCVService/pull/874/files#diff-3ad3a0b10d1284c012cbede5d466b5f325e911fd78c1e74e8686ff54dd9e9903

app/Registration.php
charlesstrange2: :+1:

resources/views/store/index_registration.blade.php
charlesstrange2: Possibly OOS in this case - it's an odd little dropdown box which switches "mode".
I'll card it for tech debt.

resources/views/store/index_registration.blade.php
charlesstrange2: better even!

resources/views/store/index_registration.blade.php
charlesstrange2: sadly, i'm just being incosistent :-/

app/Http/Controllers/Store/SessionController.php
charlesstrange2: Turns out that back() relies on a referrer header which pentest tools like Zap identify as a potential vector for insafe page redirects.

I half remembered that url()->previous() was involved in mitigating it - turns out in another project we'd overridden the base previous() method and stuck a bunch of guards in there... but that's not the case here.

this now amended with a shoter version of that hack.

tests/Feature/Store/RegistrationControllerFuzzyTest.php
```diff
@@ -0,0 +1,226 @@
...
+        $this->seeInElementAtPos('td.pri_carer', 'Smith Alice', 1);
+    }
+
+    // ── Pagination redirect ───────────────────────────────────────────────────
+
+    public function testFuzzyRedirectsToLastPageWhenRequestedPageExceedsLastPage(): void
+    {
+        $centre = factory(Centre::class)->create();
+        $user   = $this->userInCentre($centre);
+        $regs   = factory(Registration::class, 3)->create(['centre_id' => $centre->id]);
```
colinstewart-dxw: At the moment, the test appears to verify that a redirect occurs. Since it checks for page 1, it's not definite that it's hitting the last page. For example, it could be hitting the first page by mistake. The source at the moment looks correct, but it doesn't appear that the test is definitely verifying the behaviour going forward.

If that's correct, should this test be creating something like 30 registrations so there are at least 4 pages? Then it can test that a request for page 99 returns page 3, ensuring that it's definitely the last page, not the first or second (in case of off-by-one).

tests/Feature/Store/SearchPageTest.php
```diff
@@ -7,379 +7,356 @@
...
-use Tests\StoreTestCase;
 use Illuminate\Foundation\Testing\RefreshDatabase;
+use Tests\StoreTestCase;
 use URL;
 
 class SearchPageTest extends StoreTestCase
 {
     use RefreshDatabase;
 
+    // ── Helpers ───────────────────────────────────────────────────────────────
```
colinstewart-dxw, line 18: Probably want to add these to a list for abstracting to a parent class at some stage in the future.

tests/Feature/Store/SearchPageTest.php
```diff
@@ -7,379 +7,356 @@
...
-        ]);
-        $centre2 = factory(Centre::class)->create([
-            "name" => "Dagobah"
-        ]);
-        $centre3 = factory(Centre::class)->create([
-            "name" => "Coruscant"
-        ]);
+        $centre = factory(Centre::class)->create();
+        $centreUser = $this->userInCentre($centre);
+        factory(Registration::class, 10)->create(['centre_id' => $centre->id]);
```
colinstewart-dxw: Shouldn't this create more than 10 to make sure only 10 appear in the results? Or am I misunderstanding?

tests/Feature/Store/SearchPageTest.php
```diff
@@ -7,379 +7,356 @@
...
+        $centre = factory(Centre::class)->create();
+        $user = $this->userInCentre($centre);
+        factory(Registration::class, 5)->create(['centre_id' => $centre->id]);
 
-        // Check that each user we have added has a secondary info field
-        $this->assertCount(9, $this->crawler->filter('div.secondary_info'));
+        $this->actingAs($user, 'store')
+            ->visit(URL::route('store.registration.index', ['page' => '99']));
+
+        $this->seePageIs(URL::route('store.registration.index', ['page' => '1']));
```
colinstewart-dxw: Same question as earlier in relation to whether 30 registrations should be created so page 3 is returned and verifies redirection to the last page.

tests/Unit/Models/RegistrationModelTest.php
```diff
@@ -50,4 +79,186 @@ public function testItCanReturnRegistrationsOnlyForActiveFamilies(): void
...
+        // The factory creates one carer. Capture it and add a second with a higher id.
+        $reg->family->carers()->orderBy('id')->first()->update(['name' => 'Primary Carer']);
+        Carer::create(['name' => 'Secondary Carer', 'family_id' => $reg->family_id]);
+
+        $result = $this->baseQuery($centre)
+            ->select('registrations.*', 'carers.name as carer_name')
+            ->where('registrations.id', $reg->id)
+            ->first();
+
+        $this->assertEquals('Primary Carer', $result->carer_name);
```
colinstewart-dxw:
```suggestion
        $this->assertSame('Primary Carer', $result->carer_name);
```
A few other cases in this file, but of course not a blocker.

tests/Unit/Models/RegistrationModelTest.php
```diff
@@ -50,4 +79,186 @@ public function testItCanReturnRegistrationsOnlyForActiveFamilies(): void
...
+
+    public function testWithPrimaryCarerReturnsOneRowPerRegistrationRegardlessOfCarerCount(): void
+    {
+        $centre = factory(Centre::class)->create();
+
+        // Each registration gets a second carer; without the MIN(id) grouping
+        // a naive join would double the row count.
+        $regs = factory(Registration::class, 3)->create(['centre_id' => $centre->id]);
+        foreach ($regs as $reg) {
+            Carer::create(['name' => 'Extra Carer', 'family_id' => $reg->family_id]);
```
colinstewart-dxw: This depends on the factory producing an initial carer. Would it be better to add two extra carers in this loop instead of just one to cover possible changes to the factory in future?

colinstewart-dxw (approved):
That's all for this one. No strict blockers but you may still want to make some updates based on feedback. Pre-approving since any updates are likely to be relatively small and clean.

tests/Feature/Store/RegistrationControllerFuzzyTest.php
charlesstrange2: yep, amended

tests/Feature/Store/SearchPageTest.php
charlesstrange2: yep, amended

tests/Feature/Store/SearchPageTest.php
charlesstrange2: yep, amended

PR #874, opened 2026-04-20
1.20/2252 create a new collection confirmation control eee vvv part 3

charlesstrange2:
The goal here is to 
- streamline the process for _some_ centres who hand out vouchers and then walk to their own pantry and recoup them immediately as a "trader"
- we'll do this by letting them 
  a) pull from a pool of vouchers so they don't need to put in the first and last numbers.
  b) combine the handout and recollect function the system needs as a single event. 

## StoreBundleRequest
- can now cope with taking a voucher quantity OR the usual start/end combo

## Pickup and Collect blades
- have had some UX improvements
- are very similar, except Collect provides a Trader id, for the collection to pretend to be.
- (this trader id was provided by the centre having been set to collecting, and a listener on that event check/making at least one suitable trader)

## FormRequests
- form requests exist to validate the pickup and collect process
- there was an unneccesary check on vouchers in one request which was removed

## AppServiceProvider
- has a gate to prevent users not in centres performing actions they shouldn't
- (this will be used in later)

## Transition processor
This is code from the Trader facing app that does the normal job of transitioning the vouchers between states.
- it's now capable of handling calls from the centre user side that might provide a predefined set of vouchers, rather than having to resolve them from codes
- it had some utility functions added to say if there were errors and what codes it affected.

## Voucher Model
- got the ability to basically pick the next few un-bundled vouchers, so we don't have to specify vouchers by hand

## Bundle controller
  - has an function for simultaneously capturing handout and collection in a similar fashion to traders collecting vouchers.
  - this permits a centre user to act as a trader, and avoids actually handing vouchers to a family.
  - has had it's handout logic abstracted a bit so we can then reuse it for the "collecting" logic, as the system requires vouchers to be handed out before they can be collected in again. 

# On the way -

## Tests
- lots of tests :grimacing: 

## Statetoken
- got modernised, mostly to use laravel to produce v4 Uuids instead of poking Ramsey/Uuid directly.

## Store routes
- got moved about a bit by the refactors

### DIFF
--- a/app/Http/Controllers/Store/BundleController.php
+++ b/app/Http/Controllers/Store/BundleController.php
@@ -20,7 +20,6 @@ use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\HtmlString;
 use Illuminate\Support\Str;
-use Request;
 use Throwable;
 
 class BundleController extends Controller
--- a/app/Http/Controllers/Store/BundleController.php
+++ b/app/Http/Controllers/Store/BundleController.php
@@ -19,6 +19,7 @@ use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\HtmlString;
 use Illuminate\Support\Str;
+use Request;
 use Throwable;
 
 class BundleController extends Controller
@@ -59,20 +60,35 @@ class BundleController extends Controller
     }
 
     /**
-     * Append a single voucher or range of vouchers to the current bundle.
+     * Append a single voucher, range of vouchers, or a quantity drawn from the pool
+     * to the current bundle.
      */
     public function addVouchersToCurrentBundle(
         StoreAppendBundleRequest $request,
         Registration $registration,
     ): RedirectResponse {
-        $voucherCodes = Voucher::generateCodeRange(
-            $request->input('start'),
-            $request->input('end'),
-        );
-
         $managerRoute = $this->managerRoute($registration);
 
-        $errors = count($voucherCodes) <= config('arc.bundle_max_voucher_append')
+        if ($request->filled('quantity')) {
+            try {
+                $voucherCodes = Voucher::claimFromPool((int) $request->input('quantity'))
+                    ->pluck('code')
+                    ->all();
+            } catch (Throwable $e) {
+                return $this->redirectAfterRequest(
+                    ['pool' => $e->getMessage()],
+                    $managerRoute,
+                    $managerRoute,
+                );
+            }
+        } else {
+            $voucherCodes = Voucher::generateCodeRange(
+                $request->input('start'),
+                $request->input('end'),
+            );
+        }
+
+        $errors = (count($voucherCodes) <= config('arc.bundle_max_voucher_append'))
             ? $registration->currentBundle()->addVouchers($voucherCodes)
             : ['append' => count($voucherCodes)];
 
@@ -289,5 +305,12 @@ class BundleController extends Controller
             Str::plural('voucher', $count),
             $name,
         );
+    }
+
+    /**
+     *
+     */
+    public function requestPayment(Request $request) {
+        //
     }
 }
--- a/app/Http/Controllers/Store/BundleController.php
+++ b/app/Http/Controllers/Store/BundleController.php
@@ -106,14 +106,12 @@ class BundleController extends Controller
     {
         $managerRoute = $this->managerRoute($registration);
         $bundle = $registration->currentBundle();
-        $errors = [];
 
-        if ($request->filled(['collected_at', 'collected_by', 'collected_on'])) {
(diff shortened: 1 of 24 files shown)


app/Http/Controllers/Store/BundleController.php
```diff
@@ -59,66 +64,100 @@ public function create(Registration $registration): View
...
+                $errors = $this->attemptDisbursal(
+                    $bundle,
+                    // Should be here, due to form request validation
+                    $request->only(['collected_at', 'collected_by', 'collected_on'])
                 );
 
-                if (empty($errors)) {
-                    $successRoute = route('store.registration.index');
+                if (!empty($errors)) {
+                    throw new RuntimeException('disbursal errors');
```
colinstewart-dxw, line 138: Would it be handy to append the errors (even just the keys) so the log has additional context on which disbursal errors occurred?

app/Services/TransitionProcessor.php
```diff
@@ -311,4 +329,24 @@ public function constructResponseMessage(): array
             ]),
         ];
     }
+
+    /**
+     * Works out if we had any type of voucher transition failure
+     */
+    public function hasFailures(): bool
+    {
+        return (bool) Arr::first(
```
colinstewart-dxw, line 338: Could maybe just use `array_filter()` if you want (I think?). No big deal either way.
```php
return (bool) array_filter(Arr::except($this->responses, 'success_add'));
```

resources/views/store/voucher-manager/collect.blade.php
```diff
@@ -0,0 +1,108 @@
...
+
+    <div>
+        <p>There's <span class="number-circle">{{ $vouchers_amount }}</span> @choice('{1}voucher|[0,2,*]vouchers', $vouchers_amount) waiting for this {{ $programme === 0 ? 'family' : 'household' }}</p>
+    </div>
+
+    <form
+        method="POST"
+        action="{{ route('store.registration.vouchers.transitions.collect', ['registration' => $registration->id]) }}"
+    >
+        @method('PUT')
```
colinstewart-dxw, line 16: The route in `routes/store.php` has `Route::post()`. May need an update there to `Route::put()`.

routes/store.php
```diff
@@ -118,6 +118,13 @@
         Route::post('/registrations/{registration}/vouchers', [BundleController::class, 'addVouchersToCurrentBundle'])
             ->name('store.registration.vouchers.post')
             ->whereNumber('registration');
+
+        Route::post(
```
colinstewart-dxw: Just pairing this with the earlier comment regarding the form using `PUT` but the route being `POST`.

tests/Unit/Controllers/Store/BundleControllerTest.php
```diff
@@ -64,658 +58,772 @@ protected function setUp(): void
...
+     * Handles both plain strings and HtmlString instances (serialised with an 'html' key).
+     *
+     * @param array<int, string|array<string, string>> $errorMessages
+     */
+    private function hasMatchingErrorMessage(array $errorMessages, string $regex): bool
+    {
+        foreach ($errorMessages as $error) {
+            $string = is_array($error) && array_key_exists('html', $error)
+                ? $error['html']
+                : (string)$error;
```
colinstewart-dxw: Is there a situation where `$error` will be an array without the `html` key, which will then pass it to `(string)$error`?

tests/Unit/FormRequests/StoreAppendBundleRequestTest.php
```diff
@@ -0,0 +1,162 @@
...
+        $errors = session('errors')->get($field);
+
+        $this->assertNotEmpty($errors, "Expected validation errors for field '{$field}'");
+        $this->assertContains($expectedMessage, $errors);
+
+        $this->followRedirects()
+            ->seePageIs($route)
+            ->assertResponseStatus(200);
+    }
+}
```
colinstewart-dxw, line 198: No test that valid input is accepted as with other request test classes, but could always be added in future if it's wanted.

colinstewart-dxw (approved):
No real blockers, just a few items for consideration.

Pre-approving as any updates are likely to be small and clean.

app/Http/Controllers/Store/BundleController.php
charlesstrange2: I _suspect_ it's only ever going to be a database flub from disburseBundle below, and that logs as well as throws, and i reckon Laravel itself will stack out intp the logs as well.

app/Services/TransitionProcessor.php
charlesstrange2: this got moved and modified in https://github.com/neontribe/ARCVService/pull/875/files#diff-dfad123451d2fac9f2c55a3bcd1e84408c55916fbf8510f364d0ded31c1e76bc

It's largely the same there, i put the first in to force an early return.

